PR #29, opened 2022-06-29
Add Allowance/Approval methods to HTS precompile

### LINKED ISSUE #6: Update solidity file to match services implementation
### Problem

Additional methods were added to the HederaTokenService precompile flow.
However, these were not captured in this repo

### Solution

Update the solidity files to match the services usage 

### Alternatives

_No response_

shemnon:
**Description**:
Add the missing Allowance/Approval methods that can be accessed via
the ERC-20 proxy that is not available in the HTS precompile.

**Related issue(s)**:

Fixes #6

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hts-precompile/HederaTokenService.sol
+++ b/hts-precompile/HederaTokenService.sol
@@ -209,6 +209,102 @@ abstract contract HederaTokenService is HederaResponseCodes {
         (responseCode, tokenAddress) = success ? abi.decode(result, (int32, address)) : (HederaResponseCodes.UNKNOWN, address(0));
     }
 
+    /// Allows spender to withdraw from your account multiple times, up to the value amount. If this function is called 
+    /// again it overwrites the current allowance with value.
+    /// Only Applicable to Fungible Tokens
+    /// @param token The hedera token address to approve
+    /// @param spender the account authorized to spend
+    /// @param amount the amount of tokens authorized to spend.
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    function approve(address token, address spender, uint256 amount) internal returns (int responseCode)
+    {
+        (bool success, bytes memory result) = precompileAddress.call(
+            abi.encodeWithSelector(IHederaTokenService.approve.selector,
+            token, spender, amount));
+        responseCode = success ? abi.decode(result, (int32)) : HederaResponseCodes.UNKNOWN;
+    }
+
+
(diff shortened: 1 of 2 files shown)


hts-precompile/IHederaTokenService.sol
```diff
@@ -392,4 +392,55 @@ interface IHederaTokenService {
     /// @param serialNumber The serial number of the NFT to transfer.
     function transferNFT(address token,  address sender, address recipient, int64 serialNumber) external
         returns (int responseCode);
+
+    /// Allows spender to withdraw from your account multiple times, up to the value amount. If this function is called 
+    /// again it overwrites the current allowance with value.
+    /// Only Applicable to Fungible Tokens
+    /// @param token The hedera token address to approve
+    /// @param spender the account authorized to spend
+    /// @param amount the amount of tokens authorized to spend.
```
Nana-EC, line 401: nit: can we add the unit for this and other amounts. Devs may not know if its a gas unit or an hbar unit

hts-precompile/HederaTokenService.sol
```diff
@@ -209,6 +209,102 @@ abstract contract HederaTokenService is HederaResponseCodes {
...
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    function approve(address token, address spender, uint256 amount) internal returns (int responseCode)
+    {
+        (bool success, bytes memory result) = precompileAddress.call(
+            abi.encodeWithSelector(IHederaTokenService.approve.selector,
+            token, spender, amount));
+        responseCode = success ? abi.decode(result, (int32)) : HederaResponseCodes.UNKNOWN;
+    }
+
+
```
Nana-EC, line 227: nit: extra line

Nana-EC (changes requested):
Looks good.
If you don't mind can you add some nits that clarify account is account address ~and what unit the amounts are~.
It may or may not be clear depending on familiarity with Hedera and which side you're coming from so better to be explicit

hts-precompile/IHederaTokenService.sol
Nana-EC: Ignore. Mixed items up

Nana-EC (dismissed):
Unblocking since now just nits

hts-precompile/HederaTokenService.sol
shemnon: done

Nana-EC (approved):

Issue #50, opened 2022-08-25
Update KeyHelper w get key methods to test HIP 514 improvements


Nana-EC:
### Problem

KeyHelper.sol currently doesn't have the methods added in [relay tests](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/contracts/KeyHelper.sol) to be able to support easy testing 

### Solution

Update KeyHelper.sol with methods similar to [relay tests](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/contracts/KeyHelper.sol)
Note concerns highlighted by #48

Issue #48, opened 2022-08-24
Improve new KeyHelper interface


Sean-Tedrow-LB:
### Problem

I'm under the impression that the `KeyHelper.sol` file defined [here](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/contracts/KeyHelper.sol) is going to be copied to this repo.

With that updated `KeyHelper`, creating a key looks like this: `getSingleKey(0, 6, 1, bytes(""))`

In this code snippet, those three numbers, `0`, `6`, and `1` are magic numbers, where `0` indicates that we want to set the admin key, `6` indicates that we also want to set the pause key, and `1` indicates that we want the account key that signed this execute transaction to be used for both.

When you read the code snippet `getSingleKey(0, 6, 1, bytes(""))` on its own, it's very difficult to read and understand

### Solution

This interface would be greatly improved by providing constants for these magic numbers.

I'm hoping that this was always the plan and that I'm just being paranoid, but it seems like we're pretty close to release, so I just want to make sure that my concerns are documented.

PR #53, opened 2022-08-27
Add Token Managment wrapper contract

### LINKED ISSUE #52: Import Relay BaseHTS as TokenManagementContract
### Problem

The relay repo added a [BaseHTS.sol](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/contracts/BaseHTS.sol) contract that was helpful as an example that uses the wrapper contract for HIP 514 features.
This repo has no such wrapper contract with details

### Solution

Copy over the [logic](https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/server/tests/contracts/BaseHTS.sol) into a `TokenManagementContract.sol` file 

### Alternatives

_No response_

georgi-l95:
Signed-off-by: <georgi-l95 glazarov95@gmail.com>

**Description**:
Adds wrapper contract for HIP 514 features.

**Related issue(s)**:

Fixes #52 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hts-precompile/TokenManagementContract.sol
+++ b/hts-precompile/TokenManagementContract.sol
@@ -14,28 +14,6 @@ contract TokenManagementContract is FeeHelper {
     uint decimals = 8;
     bool freezeDefaultStatus = false;
 
-    event CreatedToken(address tokenAddress);
-    event DefaultFreezeStatusChanged(bool freezeStatus);
-    event AllowanceValue(uint256 amount);
-    event ResponseCode(int responseCode);
-    event ApprovedAddress(address approved);
-    event Approved(bool approved);
-    event FungibleTokenInfo(IHederaTokenService.FungibleTokenInfo tokenInfo);
-    event TokenInfo(IHederaTokenService.TokenInfo tokenInfo);
-    event NonFungibleTokenInfo(IHederaTokenService.NonFungibleTokenInfo tokenInfo);
-    event MintedToken(uint64 newTotalSupply, int64[] serialNumbers);
-    event Frozen(bool frozen);
-    event PausedToken(bool paused);
-    event UnpausedToken(bool unpaused);
-    event TokenCustomFees(IHederaTokenService.FixedFee[] fixedFees, IHederaTokenService.FractionalFee[] fractionalFees, IHederaTokenService.RoyaltyFee[] royaltyFees);
-    event TokenDefaultFreezeStatus(bool defaultFreezeStatus);
-    event TokenDefaultKycStatus(bool defaultKycStatus);
-    event KycGranted(bool kycGranted);
-    event IsToken(bool isToken);
-    event TokenType(int32 tokenType);
-    event TokenExpiryInfo(IHederaTokenService.Expiry expiryInfo);
-    event TokenKey(IHederaTokenService.KeyValue key);
-
     function createFungibleTokenPublic(
         address treasury
     ) public payable {
@@ -59,8 +37,6 @@ contract TokenManagementContract is FeeHelper {
         if (responseCode != HederaResponseCodes.SUCCESS) {
             revert ();
         }
-
-        emit CreatedToken(tokenAddress);
     }
 
     function createNonFungibleTokenPublic(
@@ -87,8 +63,6 @@ contract TokenManagementContract is FeeHelper {
         if (responseCode != HederaResponseCodes.SUCCESS) {
             revert ();
         }
-
-        emit CreatedToken(tokenAddress);
     }
 
     function createFungibleTokenWithCustomFeesPublic(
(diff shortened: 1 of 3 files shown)


hts-precompile/TokenManagementContract.sol
```diff
@@ -0,0 +1,491 @@
...
+
+    string name = "tokenName";
+    string symbol = "tokenSymbol";
+    string memo = "memo";
+    uint initialTotalSupply = 1000;
+    uint32 maxSupply = 1000;
+    uint decimals = 8;
+    bool freezeDefaultStatus = false;
+
+    event CreatedToken(address tokenAddress);
```
shemnon, line 17: Wrapper contracts should be simple pass-through contracts.  Having the contract emit these events while not emitting the same events via a direct call results in a bad user experience.

shemnon (changes requested):

hts-precompile/TokenManagementContract.sol
georgi-l95: Yes, you're right. I've removed those events.

hts-precompile/TokenManagementContract.sol
```diff
@@ -0,0 +1,388 @@
...
+        IHederaTokenService.AccountAmount[] memory transfers = new IHederaTokenService.AccountAmount[](2);
+        transfers[0] = accountAmountNegative;
+        transfers[1] = accountAmountPositive;
+
+        IHederaTokenService.TokenTransferList memory tokenTransfer =
+        IHederaTokenService.TokenTransferList(token, transfers, nftTransfers);
+        IHederaTokenService.TokenTransferList[] memory tokenTransferList = new IHederaTokenService.TokenTransferList[](1);
+        tokenTransferList[0] = tokenTransfer;
+
+        responseCode = HederaTokenService.cryptoTransfer(tokenTransferList);
```
Nana-EC, line 141: Why does this use `cryptoTransfer ` instead of `transferToken`?

hts-precompile/TokenManagementContract.sol
```diff
@@ -0,0 +1,388 @@
...
+    uint initialTotalSupply = 1000;
+    uint32 maxSupply = 1000;
+    uint decimals = 8;
+    bool freezeDefaultStatus = false;
+
+    function createFungibleTokenPublic(
+        address treasury
+    ) public payable {
+        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](4);
+        keys[0] = getSingleKey(0, 6, 1, bytes(""));
```
Nana-EC, line 21: Not in this PR but we should come back to make `getSingleKey` more intuitive.
I believe there's a ticket for this #48

Nana-EC (changes requested):
Looks good.
Just on method mixup or clarification

hts-precompile/TokenManagementContract.sol
georgi-l95: This function is a wrapper for `cryptoTransfer`, which accepts only `TokenTransferList` structure as parameter. Maybe it would be better to rename this function to `cryptoTransferPublic` and add another wrapper for `transferToken` and `transferNFT` ?

hts-precompile/TokenManagementContract.sol
georgi-l95: Yes, I was planning to revisit this in #50 to replace those magic number together with the new `KeyHelper`.

hts-precompile/TokenManagementContract.sol
Nana-EC: Yes, let's do the the latter suggestion.
Please also capture this same change in the relay repo.
Either yourself in a follow up PR or capture the changes in a ticket for the relay

georgi-l95: Okay, I have a draft PR in the relay about HTS Precompile tests, waiting for 0.30 tag for services. I can apply these changes there.

Nana-EC (approved):

hts-precompile/TokenManagementContract.sol
```diff
@@ -0,0 +1,416 @@
...
+        }
+    }
+
+    function updateTokenKeysPublic(address token, IHederaTokenService.TokenKey[] memory keys) public returns (int64 responseCode) {
+        (responseCode, keys) = HederaTokenService.updateTokenKeys(token, keys);
+
+        if (responseCode != HederaResponseCodes.SUCCESS) {
+            revert();
+        }
+    }
```
shemnon, line 415: These are all just renamed copies of functions we have in [SafeHTS.sol](https://github.com/hashgraph/hedera-smart-contracts/blob/main/safe-hts-precompile/SafeHTS.sol) - why do we need another set?

hts-precompile/TokenManagementContract.sol
```diff
@@ -0,0 +1,416 @@
...
+        IHederaTokenService.TokenTransferList memory tokenTransfer =
+        IHederaTokenService.TokenTransferList(token, transfers, nftTransfers);
+        IHederaTokenService.TokenTransferList[] memory tokenTransferList = new IHederaTokenService.TokenTransferList[](1);
+        tokenTransferList[0] = tokenTransfer;
+
+        responseCode = HederaTokenService.cryptoTransfer(tokenTransferList);
+        if (responseCode != HederaResponseCodes.SUCCESS) {
+            revert();
+        }
+    }
```
shemnon, line 145: These methods go above and beyond what we've done in the other solidity code in this repo, which has all been plain passthrough. We haven't implied defaults nor implied key structures, so I am hesitant to jump in and press "approve".

Nana-EC:
@georgi-l95 the confusion might be because this code may be seen as the go to approach vs an example
Can you move `TokenCreateContract.sol` and  `TokenManagementContract.sol` into an examples folder at `hts-precompile/examples`.
This way it's clear to people it's just examples

hts-precompile/TokenManagementContract.sol
georgi-l95: Moved the contracts to example folder.

Nana-EC (approved):
LG
Note there might be a conflict between this PR and the npm package.
Subfolders here are still desired though

georgi-l95:
I don't believe that there are going to be any conflicts. If there are, I'll resolve them.

georgi-l95:
This contract overlap with https://github.com/hashgraph/hedera-smart-contracts/pull/74

Issue #42, opened 2022-08-05
Add test support for the SafeHTS library


stoqnkpL:
https://github.com/hashgraph/hedera-smart-contracts/pull/41 adds the SafeHTS library, but some framework for testing it should be created.

PR #57, opened 2022-08-31
042 safehts test support

MiroslavGatsanoga:
**Description**:

**Related issue(s)**:

Fixes [#42](https://github.com/hashgraph/hedera-smart-contracts/issues/42)

**Notes for reviewer**:
For first time use run `npm install`. 
To compile all contracts run `npx hardhat compile`.
To run tests first setup and run [hedera local node](https://github.com/hashgraph/hedera-local-node/tree/dev) and then run `npx hardhat test`.

**Checklist**


contracts/safe-hts-precompile/SafeOperations.sol
```diff
@@ -6,6 +6,9 @@ import "./SafeHTS.sol";
 
 contract SafeOperations {
 
+    event TokenCreated(address tokenAddress);
+    event TokenInfoEvent(address token, IHederaTokenService.TokenInfo tokenInfo);
```
georg-getz, line 10: nit: both address params should probably have the same name, otherwise it kind of implies that `token` is somehow different from `tokenAddress`

Nana-EC:
Is this PR also addressing #7 ?
If so let's do that first and then address #42 in a separate PR

package.json
```diff
@@ -0,0 +1,8 @@
+{
```
Nana-EC, line 1: Please populate this with a structure similar to https://github.com/hashgraph/hedera-json-rpc-relay/blob/main/packages/relay/package.json

contracts/safe-hts-precompile/SafeOperations.sol
```diff
@@ -6,6 +6,9 @@ import "./SafeHTS.sol";
 
 contract SafeOperations {
 
+    event TokenCreated(address tokenAddress);
```
Nana-EC, line 9: Might it be better to move safe contract under `hts-precompile/examples/safe`

Nana-EC (changes requested):
It seems like to achieve testing you're turning this into an npm package as wanted to achieve in #7 
Please do this first either here or in a separate PR (just the structure, npm details and dirs) and then let the subsequent PR be the code changes with tests.

package.json
MiroslavGatsanoga: Opened https://github.com/hashgraph/hedera-smart-contracts/pull/61 to address this.

contracts/safe-hts-precompile/SafeOperations.sol
MiroslavGatsanoga: I think the safe contract should be located close to the SafeHTS library because it's meant to showcase its usage.
Keeping that in mind an alternative structure to the current one could be:
```
contracts/
    hts-precompile/
        safe/
            examples/
                SafeOperations.sol
            SafeHTS.sol
            HederaResponseCodes.sol
        IHederaTokenService.sol
        HederaTokenService.sol
        HederaResponseCodes.sol
    ...
``` 
or maybe placing the SafeHTS in a separate `libraries` directory i.e.
```
contracts/
    ...
libraries/
    examples/
        SafeOperations.sol
    SafeHTS.sol
    HederaResponseCodes.sol
```

Any thoughts?

Nana-EC: Let's go with the first option for now.
However, there should be code reuse. No need to repeat files like `HederaResponseCodes.sol` if there's no differnce.
Is it possible to share those files between the safe and non-hardened files?

Nana-EC (commented):
LG assuming

contracts/safe-hts-precompile/SafeOperations.sol
Nana-EC: Actually why can't we do the following? Where Safe is just the hardened version of `HederaTokenService.sol` and `TokenManage.sol`.
Basically just sub folders under examples and they contain their details and share necessary files
The only thing outside of the examples should be the files that are actually on the network
```
contracts/
    hts-precompile/
        examples/
            safe/
                SafeOperations.sol
                SafeHTS.sol
            token-create/
                HederaTokenService.sol
            token-manage/
                TokenManage.sol
            HederaResponseCodes.sol
        IHederaTokenService.sol
    ...
```

Nana-EC:
@georgi-l95 can you also take a look.
As some reordering might affect your PR

georgi-l95 (commented):
LG, I will rebase if needed.

contracts/hts-precompile/IHederaTokenService.sol
```diff
@@ -157,6 +157,18 @@ interface IHederaTokenService {
     struct TokenInfo {
         /// The hedera token;
         HederaToken hedera;
+        
+        /// The number of tokens (fungible) or serials (non-fungible) of the token
```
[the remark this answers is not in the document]
MiroslavGatsanoga: I've updated the order of the fields to match that of the tuple [here](https://github.com/hashgraph/hedera-services/blob/master/test-clients/src/main/java/com/hedera/services/bdd/suites/utils/contracts/precompile/HTSPrecompileResult.java#L451) because I was getting an error on GetTokenInfo.

shemnon, line 161: Does this align with the expected order in the HTS Precompile logic in services?  The ABI does not do symbols but matches strictly by ordinal positions.

Nana-EC: Asked the same question [above](https://github.com/hashgraph/hedera-smart-contracts/pull/57#discussion_r969559646) and @georgi-l95  claims it does

Nana-EC (approved):
LG

shemnon (dismissed):

package.json
```diff
@@ -14,5 +14,10 @@
   "bugs": {
     "url": "https://github.com/hashgraph/hedera-smart-contracts/issues"
   },
-  "homepage": "https://github.com/hashgraph/hedera-smart-contracts#readme"
+  "homepage": "https://github.com/hashgraph/hedera-smart-contracts#readme",
+  "devDependencies": {
+    "@hashgraph/hardhat-hethers": "^1.0.2",
```
Nana-EC, line 19: Q: Why do we need `hardhat-hethers` vs just hardhat?
If this is only needed to workaround the contract address after deployment then it's fine to have it but we should import hardhat explicitly and only use `@hashgraph/hethers` for the single purpose.
For this repo let's keep it to the bare minimum of local node and hardhat with needed configurations to work

Nana-EC (commented):
One clarifying question then you should be good to check this in

Nana-EC (approved):
LG

PR #59, opened 2022-09-06
update types for maxSupply to int64 and initialSupply to uint64

### LINKED ISSUE #56: [hts-precompile]`uint32` for `maxSupply` and `uint` for `initialSupply` in `IHederaTokenService`
### Description

In the [document](https://docs.hedera.com/guides/docs/hedera-api/token-service/tokencreate), it stated that `maxSuppy` supports int64. However in the HTS solidity contract, `IHederaTokenService.TokenInfo.maxSupply` is defined as `uint32`, which is too small to fit a token totalSupply. 
I did some tests on testnet.
1. I was able to create a token with maxSupply more than `uint32` with HTS JavaScript SDK.
2. I was not able to create a token with maxSupply more than `uint32` with HTS Solidity library
3. I tried to change the type `IHederaTokenService.TokenInfo.maxSupply` to `int64` and `uint64`. Neither of them worked. The tx just got reverted.

Also, in the [document](https://docs.hedera.com/guides/docs/hedera-api/token-service/tokencreate), `initialSupply` should be `uint64`, but in function `createFungibleToken`, `createFungibleTokenWithCustomFees`, `uint` is used instead of `uint64`.

(I talked Greg Scullard#5365 in discord. He said it appears to be a bug.)
(Credit to @Shungy)

### Steps to reproduce

1. use `createFungibleToken` to create a token with maxSupply more than `uint32` with HTS Solidity library. Token creation tx will get reverted
3. change the type of `IHederaTokenService.TokenInfo.maxSupply` to `int64` and `uint64`. Neither of them worked. The tx just got reverted.

### Additional context

_No response_

### Hedera network

testnet

### Version

commit 7bc4b3b27fef3ab38cb349b3ef5c9270009617cd

### Operating system

macOS

lukelee-sl:

**Description**:
Update the types for maxSupply and initialSupply to match the types in the protobuf specification.

**Related issue(s)**:

Fixes #56 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/hts-precompile/IHederaTokenService.sol
+++ b/hts-precompile/IHederaTokenService.sol
@@ -140,7 +140,7 @@ interface IHederaTokenService {
         // IWA Compatibility. Depends on TokenSupplyType. For tokens of type FUNGIBLE_COMMON - the
         // maximum number of tokens that can be in circulation. For tokens of type NON_FUNGIBLE_UNIQUE -
         // the maximum number of NFTs (serial numbers) that can be minted. This field can never be changed!
-        uint32 maxSupply;
+        int64 maxSupply;
 
         // The default Freeze status (frozen or unfrozen) of Hedera accounts relative to this token. If
         // true, an account must be unfrozen before it can receive the token
@@ -395,7 +395,7 @@ interface IHederaTokenService {
     /// @return tokenAddress the created token's address
     function createFungibleToken(
         HederaToken memory token,
-        uint initialTotalSupply,
+        uint64 initialTotalSupply,
         uint decimals
     ) external payable returns (int64 responseCode, address tokenAddress);
 
@@ -410,7 +410,7 @@ interface IHederaTokenService {
     /// @return tokenAddress the created token's address
     function createFungibleTokenWithCustomFees(
         HederaToken memory token,
-        uint initialTotalSupply,
+        uint64 initialTotalSupply,
         uint decimals,
         FixedFee[] memory fixedFees,
         FractionalFee[] memory fractionalFees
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):
LG

hts-precompile/IHederaTokenService.sol
```diff
@@ -395,7 +395,7 @@ interface IHederaTokenService {
     /// @return tokenAddress the created token's address
     function createFungibleToken(
         HederaToken memory token,
-        uint initialTotalSupply,
+        uint64 initialTotalSupply,
         uint decimals
```
shemnon, line 399: decimals is `uint32` - https://github.com/hashgraph/hedera-protobufs/blob/d5eaa07c41999fe8e6611b552d50373f8a809a07/services/token_create.proto#L92

How far the rabbit hole do we go?

lukelee-sl: It makes sense to me to fix it all while I'm here.  There should be no difference in the services code except the hash that is generated.

Nana-EC (approved):

hts-precompile/IHederaTokenService.sol
dimitar-dinev: Alongside the hash update, I think we can also [delete the `initSupply` and `decimals` checks in TokenCreatePrecompile](https://github.com/hashgraph/hedera-services/blob/master/hedera-node/src/main/java/com/hedera/services/store/contracts/precompile/impl/TokenCreatePrecompile.java#L284-L297), since they will no longer be needed after this type fix.

lukelee-sl: @dimitar-dinev I believe we might want to keep the checks as the we will still be supporting the old version of these types.

PR #81, opened 2022-10-04
Enhance support for atomic crypto transfers

### LINKED ISSUE #69: Add atomic crypto transfer HTS precompile support
### Problem

The `IHederaTokenService` adds HTS support to precompiles
However, since Hedera combines crypto and token transfers we missed the atomic form along with allowance support

### Solution

To close off the HTS precompile story, design and add support for atomic transfers of HBAR and Tokens, including allowance amounts

- Update `AccountAmount` struct to support `isApproval`
  ```
    struct AccountAmount {
        ...

        // If true then the transfer is expected to be an approved allowance and the
        // accountID is expected to be the owner. The default is false (omitted).
        bool isApproval;
    }
  ```
  - This creates a breaking change in the `cryptoTransfer` function hash since the `AccountAmount` and `TokenTransferList` struct hash changes and will require network support of both versions
- Update `NftTransfer` struct to support `isApproval`
  ```
    struct NftTransfer {
        ...

        // If true then the transfer is expected to be an approved allowance and the
        // accountID is expected to be the owner. The default is false (omitted).
        bool isApproval;
    }
  ```
  - This creates a breaking change in the `cryptoTransfer` function hash since the `NftTransfer` and `TokenTransferList` struct hash change and will require network support of both versions
- Add a new struct for `TransferList` which is composed of `AccountAmount[] transfers`
  ```
    struct TransferList {
        // Applicable to HBAR crypto currency. Multiple list of AccountAmounts, each of which
        // has an account, amount and isApproval flag.
        AccountAmount[] transfers;
    }
  ```
- overload `cryptoTransfer` with `cryptoTransfer(TransferList)`
- overload `cryptoTransfer` with `cryptoTransfer(TransferList, TokenTransferList)` (maybe TransferList only option is unnecessary with this option as one may always set TokenTransferList as empty for HBAR only transfer)

These present breaking options.
Given how early HTS precompile usage is this may be acceptable now with versioning support on the network and pointing documentation to reference old but highlight the new.

### Alternatives
Non breaking options
- alternative `AccountAmount` struct e.g. `ApprovalAccountAmount` that inherits `AccountAmount` and is specifically for allowances and used in `TransferList`. This would then require an `ApprovalTokenTransferList` and an `ApprovalNftTransfer`. This is non breaking but provides a naming difference when compared to HAPI APIs and additional structs.
- Add `cryptoTransferHbar(TransferList)`
- Add `cryptoTransferHbarAndTokens(TransferList, TokenTransferList)` (maybe this option is less desirable since it mixes an apporval struct with a non approvable struct)
- Add `cryptoTransferHbarAndTokens(TransferList, ApprovalTokenTransferList)` (maybe TransferList only option is unnecessary with this option as one may always set ApprovalTokenTransferList as empty for HBAR only transfer)

_No response_

lukelee-sl:
update structs and function signatures to support atomic crypto transfers

**Description**:
Update structs to add boolean for allowance.  Update the function signature to also pass in changes to hbar balances in addition to the existing token transfers.

Fixes #69 


contracts/safe-hts-precompile/SafeOperations.sol
```diff
@@ -42,8 +42,8 @@ contract SafeOperations {
         SafeHTS.safeTransferNFT(token, sender, receiver, serialNum);
     }
 
-    function safeCryptoTransfer(IHederaTokenService.TokenTransferList[] memory tokenTransfers) external {
-        SafeHTS.safeCryptoTransfer(tokenTransfers);
+    function safeCryptoTransfer(IHederaTokenService.TransferList[] memory transferList, IHederaTokenService.TokenTransferList[] memory tokenTransfers) external {
```
Nana-EC, line 45: Please add test coverage in [SafeHTS tests](https://github.com/hashgraph/hedera-smart-contracts/blob/main/test/safe-hts-precompile/SafeHTS.js)

Nana-EC: Add a `event CryptoTransfer` that gets emitted after with the response details and you can check for that in the test

Nana-EC (changes requested):
Looking good.
Let's get into the practice of adding the solidity test coverage where we can not that we have the setup.
I've noted where you can add a test

contracts/safe-hts-precompile/SafeOperations.sol
Nana-EC: Actually, I realize now there's a limit to how much we can test at this point without local node also having the services node that supports the functionality.
Had this been a normal contract we wouldn't be tied.
Update the contract but it's okay if we're not able to get a full test right now

PR #88, opened 2022-10-07
Enhance support for atomic crypto transfers

### LINKED ISSUE #69: Add atomic crypto transfer HTS precompile support

lukelee-sl:

**Description**:
Update structs to add boolean for allowance. Update the function signature to also pass in changes to hbar balances in addition to the existing token transfers.

Fixes #69 

### DIFF
--- a/contracts/hts-precompile/HederaTokenService.sol
+++ b/contracts/hts-precompile/HederaTokenService.sol
@@ -18,14 +18,16 @@ abstract contract HederaTokenService is HederaResponseCodes {
         _;
     }
 
-    /// Initiates a Token Transfer
+    /// Performs transfers among combinations of tokens and hbars
+    /// @param transferList the list of hbar transfers to do
     /// @param tokenTransfers the list of transfers to do
     /// @return responseCode The response code for the status of the request. SUCCESS is 22.
-    function cryptoTransfer(IHederaTokenService.TokenTransferList[] memory tokenTransfers) internal
+    /// @custom:version 0.3.0 the signature of the previous version was cryptoTransfer(TokenTransferList[] memory tokenTransfers)
+    function cryptoTransfer(IHederaTokenService.TransferList memory transferList, IHederaTokenService.TokenTransferList[] memory tokenTransfers) internal
     returns (int responseCode)
     {
         (bool success, bytes memory result) = precompileAddress.call(
-            abi.encodeWithSelector(IHederaTokenService.cryptoTransfer.selector, tokenTransfers));
+            abi.encodeWithSelector(IHederaTokenService.cryptoTransfer.selector, transferList, tokenTransfers));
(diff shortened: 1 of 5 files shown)


Nana-EC (approved):
LG
Add an issue to enable this test support post availability on the services side

PR #143, opened 2022-12-05
Add Compare Test for HAPI vs Precompile Token

### LINKED ISSUE #141: Add tests for HTS token created with precompile call vs created with HAPI TX
### Problem

We want to ensure that when the token is created via HAPI or when it is created via HTS precompile both work and give the same result.

### Solution

We should have test case do the same HTS flow via HAPI and system contract.

### Alternatives

_No response_

### LINKED ISSUE #142: Add solidity-analyzer dependency to package.json
### Problem

Currently the hardhat compiler requires `solidity-analyzer` to be installed but they aren't by default

Fails with

```
An unexpected error occurred:

Error: Cannot find module '@nomicfoundation/solidity-analyzer-darwin-arm64'
...
code: 'MODULE_NOT_FOUND',
...
```

### Solution

Add `@solidity-analyzer` to the dependencies

### Alternatives

_No response_

georgi-l95:
Signed-off-by: georgi-l95 <glazarov95@gmail.com>

**Description**:

- Adds test for token created with HAPI transaction versus token created via Precompile and compare their properties.

**Related issue(s)**:

Fixes #141 

**Notes for reviewer**:

**Checklist**


test/hts-precompile/utils.js
```diff
@@ -198,6 +214,28 @@ class Utils {
       expect(e.code).to.eq(code);
     }
   }
+
+  static async createClient(operatorId, operatorKey, hederaNetwork = { "127.0.0.1:50211": new AccountId(3) }, mirrorNode = "127.0.0.1:5600") {
```
Nana-EC, line 218: Was going to suggest to add a config file to pick up client details but for now maybe rename to `createLocalSDKClient` so it's explicit

Nana-EC (dismissed):
LG
I think future tickets might add some follow up actions on the tokens to see comparable transaction expectations.

@gregscullard any suggestions?

Nana-EC (approved):

PR #156, opened 2022-12-22
Add Proxy Upgradable Contract Example w HTS

### LINKED ISSUE #84: Add proxy upgrade contract example w HTS
### Problem

Hedera is able to support upgradeable contracts using the proxy method.
However, no example exists in this repo yet

### Solution

Add 
- a set of contracts example under `contracts/proxy-upgrade` that showcase how an upgradeable contract may be utilized w HTS logic e.g. very simple DEX ref.
- unit tests that show valid contract execution
- [Stretch] manual upgrade flow in a test that verifies eventual updated logic

Note there may be some dev tool limitations due to CREATE 1 logic by Hedera vs Ethereum but this should either have documentation or a work around provided.

### Alternatives

_No response_

georgi-l95:
**Description**:

Add two Proxy Upgradable UUPS Contracts and Tests:
- Counter
- Simple DEX with Hedera token

Bump Network and Mirror Node images.
Bump NodeJS version to the latest LTS - 18.12.1

**Related issue(s)**:

Fixes #84 

**Notes for reviewer**:

**Checklist**


contracts/proxy-upgrade/exchange/Exchange.sol
```diff
@@ -0,0 +1,98 @@
...
+import "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "../../hts-precompile/IHederaTokenService.sol";
+import "../../hts-precompile/HederaResponseCodes.sol";
+
+contract Exchange is OwnableUpgradeable, UUPSUpgradeable {
+    address public tokenAddress;
+    address private precompile;
```
natanasow, line 13: Let's assign the value directly here and make it constant `address constant private precompile = address(0x167);`

natanasow (approved):
LG

contracts/proxy-upgrade/exchange/Exchange.sol
georgi-l95: Done.

natanasow (dismissed):

lukelee-sl:
LG.  Are there any "negative" tests we want to consider?  For example, trying to call at the version() function before the upgrade.  I think it's fine without but wanted to just raise the question.

lukelee-sl (dismissed):

georgi-l95:
You're right, i've added negative cases.

lukelee-sl (approved):

PR #162, opened 2023-01-11
Added a warning about using the correct version of the interface

Neurone:
**Description**:

Added a warning about using the correct version of the interface.

**Related issue(s)**:

N/A

**Notes for reviewer**:

The current implementation in the `main` branch is very different from that in production. It is crucial developers use the correct interface; this is why I set the note at the beginning of the page.

**Checklist**

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Hedera Smart Contracts
 Reference library for Smart Contracts utilized by the Hedera network with supporting files and examples.
 
+Please checkout `v0.2.0` if you want to use the interface currently implemented in the Mainnet.
+
 ## Overview
 The Hedera network utilizes system contracts at a reserved contract address on the EVM to surface HAPI service functionality through EVM processed transactions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,8 @@
 # Hedera Smart Contracts
 Reference library for Smart Contracts utilized by the Hedera network with supporting files and examples.
 
+Please checkout `v0.2.0` if you want to use the interface currently implemented in the Mainnet.
```
Nana-EC, line 4: I'm not sure this is the case or comment to address it.

The current version in main is deployed to Hedera Mainnet.
Is there an issue you're noting that you believe v0.2.0 this version doesn't match.
Notably there are changes since v0.2.0, these will be capture in a v0.3.0 release

Nana-EC (changes requested):

Nana-EC:
@Neurone thanks for the note.
We'll address this through #166

PR #168, opened 2023-01-24
Multicaller tests

Ivo-Yankov:
**Description**:
Use the `Multicall3` contract from https://github.com/mds1/multicall to test calls with large input/output for pure and payable state-changing methods.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


contracts/multicaller/Multicall3.sol
```diff
@@ -0,0 +1,216 @@
+// SPDX-License-Identifier: MIT
```
david-bakin-sl, line 1: Please add either inside this file or as a adjoining _readme_ file the path to the github source including commit hash

test/multicall/Multicall.js
```diff
@@ -0,0 +1,319 @@
...
+
+const {expect} = require("chai");
+const {ethers} = require('hardhat');
+const {solidityPack} = require("ethers/lib/utils");
+const {defaultAbiCoder} = require("@ethersproject/abi");
+
+describe('Multicall3', function () {
+
+    let multicaller, receiver, reverter;
+    const REVERT_REASON_DATA = '0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000174d756c746963616c6c333a2063616c6c206661696c6564000000000000000000';
```
david-bakin-sl, line 29: Please add the "interpretation" of this hex blob (I had to go to a hex-to-ascii converter site to find out the message is 'Multicall3: call failed')

test/multicall/Multicall.js
```diff
@@ -0,0 +1,319 @@
...
+        reverter = await deployContract('Reverter');
+    });
+
+    describe('static calls with large input', async function () {
+        it('should be able to make processLongInput calls with length 10', async function () {
+            const n = 10;
+            const {callData, data} = prepareLongInputData(n, 'processLongInput');
+
+            const dataSize = getInputLengthInBytes(data);
+            expect(dataSize).to.be.eq(2660);    // data is 2.6 kb
```
david-bakin-sl, line 161: I see from here and subsequent tests that 266 is the length of the input data single element - not sure why.  Is that some kind of standard?  Computed in this test somewhere?  Use named constants here perhaps, for 10 * INPUT_ELEMENT_LENGTH or something like that?

test/multicall/Multicall.js
```diff
@@ -0,0 +1,319 @@
...
+    })
+
+    describe('static calls with large output', async function () {
+        it('should be able to make processLongOutput calls with length 13 kb', async function () {
+            const n = 10;
+            const res = await multicallProcessLongOutput(n);
+            expect(res).to.exist;
+            expect(res.length).to.eq(n);
+            const bytes = getOutputLengthInBytes(res.map(r => r.returnData));
+            expect(bytes).to.eq(13440); // 13 kb
```
david-bakin-sl, line 228: Again, here I don't know where 13440 comes from, same for 824320 in the next test.  At least a comment?

david-bakin-sl (commented):
I know _you_ know what's going on when you wrote these tests, but as a reader I found the magic numbers mysterious.  You might find them mysterious yourself next year when you're troubleshooting some reported issue! (Test names were fine though.)

test/multicall/Multicall.js
```diff
@@ -0,0 +1,328 @@
...
+const {defaultAbiCoder} = require("@ethersproject/abi");
+
+describe('Multicall3', function () {
+
+    let multicaller, receiver, reverter;
+
+    // Translates to: "Multicall3: call failed"
+    const REVERT_REASON_DATA = '0x08c379a0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000174d756c746963616c6c333a2063616c6c206661696c6564000000000000000000';
+    const RESULT_FIVE = '0x0000000000000000000000000000000000000000000000000000000000000005';
+    const INPUT_ELEMENT_LENGTH = 266;
```
david-bakin-sl, line 33: ok!  (now my lack of encoding knowledge shows up: I can account for 264 bytes = selector + 4 256-bit params, but where does the last (encoded) byte come from?)

(you could if you wanted add a comment here: selector + 4 params + ? (that last byte))

test/multicall/Multicall.js
```diff
@@ -0,0 +1,328 @@
...
+    const LONG_INPUT_ABI = 'processLongInput((uint256,uint256,uint256,uint256))';
```
david-bakin-sl, line 34: I think it is an improvement that there are now these 8 named constants

test/multicall/Multicall.js
```diff
@@ -0,0 +1,328 @@
...
+                const res = await multicallProcessLongInput(callData);
+            } catch (e) {
+                hasError = true;
+            }
+
+            expect(hasError).to.eq(true);
+        });
+
+
+        it('should be able to aggregate 11 calls to processLongInput and handles a revert', async function () {
```
david-bakin-sl, line 208: comment says 11 but constant below @211 is 10?

oh, i see, there's an internal detail of prepareLongInputData that's exposed here.  that's a bit tricky.

test/multicall/Multicall.js
```diff
@@ -0,0 +1,328 @@
...
+    })
+
+    describe('static calls with large output', async function () {
+        it('should be able to aggregate 10 calls to processLongOutput and handle 13 kb of output data', async function () {
+            const n = 10;
+            const res = await multicallProcessLongOutput(n);
+            expect(res).to.exist;
+            expect(res.length).to.eq(n);
+            const bytes = getOutputLengthInBytes(res.map(r => r.returnData));
+            expect(bytes).to.gte(13000); // 13 kb
```
david-bakin-sl, line 230: I still don't get where the 13000 or 820000 comes from - seems to be an interaction between what's going on in the contract and how multicallProcessLongOutput is building the multicall using incrementing values of `n` for each call?  Lot's of moving parts to have to know all at once ...

david-bakin-sl (approved):
Approved (though it would be nice if these new comments, esp. w.r.t. expected output length, were addressed)

test/multicall/Multicall.js
Ivo-Yankov: The numbers here are arbitrary and not that important. All of those tests are intended to test the input and output size limits of pure and non-pure methods using a multicall contract. Most of those limits are actually not intentional and they may be different of every network because of the differences in gas price.

PR #169, opened 2023-01-25
HIP-583: Ethereum Transfer TX via Precompile

### LINKED ISSUE #161: HIP-583: Ethereum Transfer TX via Precompile
### Problem

Add positive and negative test scenarios for HIP-583 with ethereum transfers via precompile.

### Solution

- HBAR
   - assert that hollow account is created and the amount of HBARs is correctly transferred via precompile
   - assert that second transfer HBARs via precompile to the hollow account is successful
   - assert that making HBAR transfer via precompile from hollow account to another and signing the transaction with the `ECDSA` key that corresponds to the EVM address of the hollow account is finalizing (in the state the key of the hollow account should be the `ECDSA` key which corresponds to the EVM address)

- Fungible token
   - assert that hollow account is created and the amount of fungible tokens is correctly transferred via precompile
   - assert that second transfer fungible tokens via precompile to the hollow account is successful
   - assert that making fungible tokens transfer via precompile from hollow account to another and signing the transaction with the `ECDSA` key that corresponds to the EVM address of the hollow account is finalizing (in the state the key of the hollow account should be the `ECDSA` key which corresponds to the EVM address)

- Non-fungible token
   - assert that hollow account is created and the amount of non-fungible token is correctly transferred via precompile
   - assert that second transfer non-fungible token via precompile to the hollow account is successful
   - assert that making non-fungible token transfer via precompile from hollow account to another and signing the transaction with the `ECDSA` key that corresponds to the EVM address of the hollow account is finalizing (in the state the key of the hollow account should be the `ECDSA` key which corresponds to the EVM address)

### Alternatives

_No response_

natanasow:
Signed-off-by: nikolay <n.atanasow94@gmail.com>

**Description**:

**Related issue(s)**:

Fixes #161 

**Notes for reviewer**:

**Checklist**


test/hip-583/HIP583.js
```diff
@@ -545,4 +545,154 @@ describe("Contract Transfer TX", function () {
...
+
+    it("should test that second transfer fungible tokens via precompile to the hollow account is successful", async function () {
+      const hollowBalanceBefore = await erc20Contract.balanceOf(tokenAddress, hollowWallet.address);
+      await tokenTransferContract.transferTokensPublic(tokenAddress, [signers[0].address, hollowWallet.address], [-amount, amount], {gasLimit: 1_000_000});
+      const hollowBalanceAfter = await erc20Contract.balanceOf(tokenAddress, hollowWallet.address);
+
+      expect(hollowBalanceAfter).to.eq(hollowBalanceBefore.add(amount));
+    });
+
+    it("should test that second transfer fungible tokens via precompile to the hollow account is successful", async function () {
```
lukelee-sl, line 612: Should this test be here?  The description is the same as the one above but the test body is the same as the one below?

Nana-EC (approved):

Issue #235, opened 2023-05-11
Update contracts.maxNumWithHapiSigsAccess to 0 and update tests


Kalina-Todorova:
Update the local node property contracts.maxNumWithHapiSigsAccess to 0 and update tests respectively

PR #240, opened 2023-05-19
Add Shanghai test support #232

### LINKED ISSUE #232: Add Shanghai test support
### Problem

In 0.38 services added support for the Shanghai EVM fork.
We should add smart contracts that ensure support

### Solution

Add smart contract tests to confirm the new Opcodes similar to [OpcodesContract.sol](https://github.com/hashgraph/hedera-services/blob/develop/hedera-node/test-clients/src/main/resource/contract/contracts/OpcodesContract/OpcodesContract.sol)

- opShl
- opShr
- opSar
- opExtCodeHash
- opPush0

More info here - https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md

### Alternatives

_No response_

attest08:
**Related issue(s)**:

Fixes #232


natanasow (dismissed):
LGTM

yiliev0 (dismissed):
LGTM

lukelee-sl:
Could we create a contract using solidity 0.8.20 that creates the push0 operation without having to use yul?

test/shanghai-opcodes/ShanghaiOpcodes.js
```diff
@@ -0,0 +1,51 @@
...
+    const hash = res.toString().slice(2)
+    expect(hash.length).to.equal(64);
+
+    expect(res).not.to.equal(ethers.constants.HashZero)
+  });
+
+  it("should be able to execute opPush0()", async function () {
+    const res = await shanghaiContract.opPush0();
+    // push0
+    expect(res).to.equal(95);
```
david-bakin-sl, line 49: Suggest: Either use _hex_ notation here and in `ShanghaiOpcodes.sol` or use decimal notation in both places. Don't mix.  Confusing.  Unless you leave a comment behind in one place or the other.

artifacts/contracts/shanghai-opcodes/ShanghaiOpcodes.sol/ShanghaiOpcodes.json
```diff
@@ -0,0 +1,115 @@
+{
```
david-bakin-sl, line 1: Since currently the solidity compilation is _not_ integrated there's no way to know that the Shanghai EVM was targetted _or_ that the `PUSH0` (among others) was generated.  Strongly suggest using the command line options to dump the targeted EVM version _and_ the assembly language into json compiler output and commit those files as well.  So then we can all understand that what claims to be tested is in fact tested.

Wouldn't hurt to put the actual solidity compiler command line in here somewhere too - perhaps as a comment in the sol file.

test/shanghai-opcodes/ShanghaiOpcodes.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it("should be able to execute opShr()", async function () {
+    const res = await shanghaiContract.opShr(2, 500);
+    // shift right
+    expect(res).to.equal(125);
+  });
+
+  it("should be able to execute opSar()", async function () {
+    const res = await shanghaiContract.opSar(1, 2);
```
david-bakin-sl, line 28: This would be a bit more convincing with a different target number than `2`.  Like the examples above.

david-bakin-sl (commented):
Suggestions.  But I would really like to see the compiler command line in here somewhere, and also compiler output proving that the right opcodes were generated.

attest08:
PR is blocked due to the following issue: https://github.com/hashgraph/hedera-smart-contracts/issues/235
Cannot be merged until the issue is resolved!

Nana-EC:
> PR is blocked due to the following issue: #235 Cannot be merged until the issue is resolved!

Additionally there's this PR https://github.com/hashgraph/hedera-services/pull/6773 which should set this for you.
We'd still want the ability to config it in local node within transactions

Nana-EC (dismissed):
LG

lukelee-sl (dismissed):
LGTM

georgi-l95 (approved):
LGTM

PR #269, opened 2023-08-01
System contract dapp playground

quiet-node:
**Description**: This PR adds an initialization of the Hedera System Contract Dapp Playground to the project
* Project setup by nextjs@13.4.12
* Configured with
     + TailwindCSS
     + ChakraUI
     + Basic SEO metadata
     + Styrene A Web fonts
     + Hedera branding colors

**Checklist**


system-contract-dapp-playground/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "system-contract-dapp-playground",
```
Nana-EC, line 2: Take a look at the package.json file in the root dir to see the format of fields and other fields to fill in

system-contract-dapp-playground/README.md
```diff
@@ -0,0 +1,43 @@
...
+### Requirements
+
+- [git](https://git-scm.com/)
+- [npm](https://www.npmjs.com/)
+- [nodejs](https://nodejs.org/en/)
+- [contributing guide](https://github.com/hashgraph/.github/blob/main/CONTRIBUTING.md#pull-requests)
+- [Hedera accounts](https://docs.hedera.com/hedera/getting-started/introduction#create-hedera-portal-profile-faucet)
+- [prettier pluggin](https://prettier.io/) (recommended)
+
+## Quickstart
```
Nana-EC, line 23: Add a `## Config` section and note all the different config files and their purposes.
There's a few more than usual so let's get them noted either here or in the next PR

system-contract-dapp-playground/src/fonts/index.ts
```diff
@@ -0,0 +1,101 @@
+import localFont from 'next/font/local';
```
Nana-EC, line 1: All `.ts` files should have a license header.
You can do this in the next PR

system-contract-dapp-playground/src/fonts/styreneA-webfont/1802-UFGULA.css
```diff
@@ -0,0 +1,240 @@
+/*
+        This font software is the property of Commercial Type.
+
+        You may not modify the font software, use it on another website, or install it on a computer.
+
+        License information is available at http://commercialtype.com/eula
+        For more information please visit Commercial Type at http://commercialtype.com or email us at info[at]commercialtype.com
+
+        Copyright (C) 2018 Schwartzco Inc.
+        License: 1802-UFGULA
```
Nana-EC, line 10: Erm you might have to remove this. We only want to use open source software. It doesn't sounds like this font package allows
Given the warning I'm not sure how it intends to be used.
What package or tool is this from?

Nana-EC (changes requested):
Nice structure.
Most comments can be addressed in a follow up PR but we should figure out the font situation

system-contract-dapp-playground/src/fonts/styreneA-webfont/1802-UFGULA.css
quiet-node: Hello @Nana-EC, thank you for the feedback. I noticed that the Styrene A font is not listed on Google Fonts, and I learned that Hedera employs Styrene A as the primary sans-serif font family. That's why I prefer to stick to this font. These files were downloaded from the Hedera branding site which can be found here https://hederabrandcentral.frontify.com/d/Tmocz52AXpLj/brand-assets#/brand-assets/typography/download-fonts

Nana-EC (approved):
LG. Can fix up other items in a follow up

PR #299, opened 2023-08-07
Add return statement to `HederaTokenService.redirectForToken()`

### LINKED ISSUE #298: Add return statement to `HederaTokenService.redirectForToken()`
### Problem

Currently, there is no return statement in `HederaTokenService.redirectForToken()`, so we can not call that contract from an external contract and get the result.

### Solution

Add a return statement with the result.

### Alternatives

_No response_

natanasow:
**Description**:

Currently, there is no return statement in HederaTokenService.redirectForToken(), so we can not call that contract from an external contract and get the result.

**Related issue(s)**:

Fixes #298 

**Notes for reviewer**:

**Checklist**


Nana-EC (changes requested):
Nice catch.
1 question

contracts/hts-precompile/IHederaTokenService.sol
```diff
@@ -797,5 +797,5 @@ interface IHederaTokenService {
 
     /// Initiates a Redirect For Token
     /// @param encodedFunctionSelector The function selector from the ERC20 interface + the bytes input for the function called
-    function redirectForToken(address token, bytes memory encodedFunctionSelector) external;
+    function redirectForToken(address token, bytes memory encodedFunctionSelector) external returns (int64 responseCode, bytes memory response);
```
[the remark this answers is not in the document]
natanasow: It was just incorrect. Nothing to change on services :sunglasses: .

Nana-EC, line 800: nit: update function comments with return details

contracts/hts-precompile/HederaTokenService.sol
```diff
@@ -660,11 +660,12 @@ abstract contract HederaTokenService {
     /// Redirect for token
     /// @param token The token address
     /// @param encodedFunctionSelector The function selector from the ERC20 interface + the bytes input for the function called
-    function redirectForToken(address token, bytes memory encodedFunctionSelector) external {
-        (bool success, bytes memory data) = precompileAddress.call(
+    function redirectForToken(address token, bytes memory encodedFunctionSelector) external returns (int responseCode, bytes memory response) {
```
Nana-EC, line 663: nit: update function comments with return

Nana-EC (approved):
LG.
Unblocking but also noted nit comments

Issue #279, opened 2023-08-04
System Contract DApp Playground: Phase 2


Nana-EC:
### Problem

Smart contract development and troubleshooting can be time consuming.
A sample and reference DApp to make it easy to test Hedera System Contract would be very useful

### Solution

Add support to the DApp to deploy all the system contracts
```[tasklist]
### System Contract DApp Playground: Phase 2
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/281
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/280
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/291
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/295
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/302
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/303
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/304
```

PR #305, opened 2023-08-08
feat: added new UI component for HTS-206 route & reworked smart contract assets (#302)

### LINKED ISSUE #302: [System Contract DApp] New UI component for HTS-206 route & rework smart contract assets
### Problem

The HTS-206 route needs a UI component. The smart contract assets currently are set up in a way to work with Nodejs module, while the React app ("esnext") also needs to access the assets. Also the assets need more client-side information.

### Solution

- Add a new UI component for HTS-206
- Reworked the smart contract assets for more client-side information and easier to access

### Alternatives

_No response_

quiet-node:
**Description**: This PR adds a new UI component for HTS-206 route and reworked smart contract assets for better information and access within React app

**Related issue(s)**: #279 

Fixes #302 

**UI update**:
image

**Checklist**

### DIFF
--- a/system-contract-dapp-playground/src/sections/hts-hip-206/index.tsx
+++ b/system-contract-dapp-playground/src/sections/hts-hip-206/index.tsx
@@ -1,0 +1,121 @@
+'use client';
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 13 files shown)


system-contract-dapp-playground/src/sections/hts-hip-206/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+          <p className="tracking-tight text-white/70">
+            These example smart contracts demonstrate how the HTS system contract is employed to
+            carry out their functions effectively. They serve as clear examples of how the HTS
+            system contract can be put to practical use.
+          </p>
+        </div>
+
+        {/* Contracts */}
+        <ol className="px-6 flex flex-col gap-9">
+          {HEDERA_SMART_CONTRACTS_ASSETS.slice(0, 4).map((contract, index) => (
```
Nana-EC, line 96: Hmm `slice(0, 4)` could turn out to be problematic.
Essentially this requires the order of assets never change in the 1st 4 elements.
If it ever did the impact would be unclear.
Is there another constant that can hold what should be displayed. Maybe it feeds into `HEDERA_SMART_CONTRACTS_ASSETS`

Nana-EC (approved):
LG.
One item of concern, not blocking but needs to be resolved in a follow up PR at least to avoid future regressions

Issue #314, opened 2023-08-10
System Contract DApp Playground: Phase 3


quiet-node:
### Problem

In [phase 2](https://github.com/hashgraph/hedera-smart-contracts/issues/279), the DApp has successfully achieved the capability to deploy the smart contracts. Phase 3's primary emphasis will shift towards refining contract interactions. This requires making the methods of the smart contracts accessible within the DApp Playground.

### Solution

Incorporate essential Solidity methods, UI components, contract-interacting APIs, and develop unit tests tailored for each individual smart contract.

```[tasklist]
### System Contract DApp Playground: Phase 3
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/316
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/322
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/321
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/319
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/318
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/317
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/315
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/340
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/368
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/371
```

PR #328, opened 2023-08-14
feat: added necessary solidity methods and its unit tests for tokenCreateCustom.sol (#315)

### LINKED ISSUE #315: [System Contract DApp] Add necessary Solidity methods and its unit tests for TokenCreateCustom.sol
### Problem

Currently, the `TokenCreateCustom.sol` contract has only one method, `createFungibeTokenPublic`, which serves the purpose of generating fungible tokens through the utilization of dynamic inputs. However, to effectively expose various functions made available by the HTS precompile contract, it becomes imperative to introduce additional methods within the `TokenCreateCustom.sol` contract.

### Solution

Add more necessary methods to TokenCreateCustom.sol contract

### Alternatives

_No response_

quiet-node:
**Description**: This PR Add more necessary methods to TokenCreateCustom.sol contract and its corresponded unit tests

**Related issue(s)**: #314

Fixes #315

**Notes for reviewer**:

**Checklist**


contracts/hts-precompile/examples/token-create/TokenCreateCustom.sol
```diff
@@ -23,7 +25,7 @@ contract TokenCreateCustomContract is HederaTokenService, ExpiryHelper, KeyHelpe
         bytes memory key
     ) public payable {
         IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](5);
-        keys[0] = getSingleKey(KeyType.PAUSE, KeyValueType.SECP256K1, key);
+        keys[0] = getSingleKey(KeyType.ADMIN, KeyType.PAUSE, KeyValueType.SECP256K1, key);
```
Nana-EC, line 28: this looks broken, should be 3 param values.
Not sure what you're trying to change

test/hts-precompile/token-create/tokenCreateCustomContract.js
```diff
@@ -0,0 +1,391 @@
...
+ * limitations under the License.
+ *
+ */
+
+const utils = require('../utils')
+const { expect } = require('chai')
+const Constants = require('../../constants')
+const { expectValidHash } = require('../assertions')
+
+describe.only('TokenCreateCustomContract Test Suite', () => {
```
Nana-EC, line 26: What's the only here for?
I think this is breaking CI

Nana-EC (changes requested):
Pointed out some breaking items

test/hts-precompile/token-create/tokenCreateCustomContract.js
quiet-node: Ahh I forgot to remove it during the clean up. Thanks for the catch!

contracts/hts-precompile/examples/token-create/TokenCreateCustom.sol
quiet-node: There are three `getSingleKey` functions within the KeyHelper.sol here is the one with 4 params https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/KeyHelper.sol#L96. 

I was trying to set up both ADMIN and PAUSE type at keys[0] as the minting and grantingKYC test cases didn't work without ADMIN type set. I got the reference from this method https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/examples/token-create/TokenCreateContract.sol#L28

Before when the ADMIN type is not set and the test suites still all worked as there is only one test utilizing the `createFungibleTokenPublic` https://github.com/hashgraph/hedera-smart-contracts/blob/main/test/hts-precompile/token-create/tokenCreateContract.js#L369, and this test case didn't require updating token keys to mint or grant token KYC.

I just push a new commit which remove the .only and all CI tests pass now

Nana-EC: I see
Q: Would it be cleaner to leave the admin key as such and add a `keys[4] = getSingleKey(KeyType.PAUSE, KeyValueType.SECP256K1, key);`?

quiet-node: Did you mean replace `keys[4] = getSingleKey(KeyType.WIPE, KeyValueType.SECP256K1, key);` with `keys[4] = getSingleKey(KeyType.PAUSE, KeyValueType.SECP256K1, key);` or you meant some thing like this 
```
        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](6);
        keys[0] = getSingleKey(KeyType.ADMIN, KeyValueType.SECP256K1, key);
        keys[1] = getSingleKey(KeyType.KYC, KeyValueType.SECP256K1, key);
        keys[2] = getSingleKey(KeyType.FREEZE, KeyValueType.SECP256K1, key);
        keys[3] = getSingleKey(KeyType.SUPPLY, KeyValueType.SECP256K1, key);
        keys[4] = getSingleKey(KeyType.WIPE, KeyValueType.SECP256K1, key);
        keys[5] = getSingleKey(KeyType.PAUSE, KeyValueType.SECP256K1, key);
```

Replacing doesn't work for minting and granting either but adding one more key to the `keys` array works.

Nana-EC: I meant adding the key to the array

quiet-node: Yes! It works like that too! Will push in the update soon!

Nana-EC (approved):

PR #343, opened 2023-08-16
test: added unit tests to fix #342

### LINKED ISSUE #342: TokenCreateCustomContract.createFungibleTokenPublic does not accept a key that's not the caller's public key
### Problem

TokenCreateCustomContract.createFungibleTokenPublic does not accept a key that's not the caller's public key

### Solution

add unit tests where keys are not caller's public key is set as the key param for the .createFungibleTokenPublic() method would fail

### Alternatives

_No response_

quiet-node:
**Description**: This PR contains two unit tests to make sure that the caller's public key needs to be set as the `key` param for the create token functions

Fixes #342

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/test/hts-precompile/token-create/tokenCreateCustomContract.js
+++ b/test/hts-precompile/token-create/tokenCreateCustomContract.js
@@ -391,7 +391,7 @@ describe('TokenCreateCustomContract Test Suite', () => {
   })
 
   describe('Key params', () => {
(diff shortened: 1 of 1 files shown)


Nana-EC (commented):
Suggestion for test name and then I think we're good

Nana-EC (approved):

Issue #326, opened 2023-08-11
[System Contract DApp - ERC20] Add contract interacting APIs and Unit Tests


quiet-node:
### Problem

The ERC20Contract requires a set of APIs to interact with contract on chain

### Solution

- Develop APIs to interact with methods from TokenCreateCustom contract
- Create unit tests to verify the accuracy of the contract's APIs.

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/330
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/332
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/333
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/335
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/336
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/337
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/338
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/339
```

PR #345, opened 2023-08-16
dapp-feat: added APIs, unit tests for tokenInformation and integrated new logic to TokenInformation UI component (#335)

### LINKED ISSUE #335: [System Contract DApp - ERC20] Add APIs & unit tests to support querying token information
### Problem

The dapp needs a set of APIs and unit tests to support querying token information from ERC20Mock contract

### Solution

- Add APIs and unit tests
- Integrated the APIs to UI

### Alternatives

_No response_

quiet-node:
**Description**: This PR adds new APIs and unit tests for token information querying feature and integrated it to the UI
 
**Related issue(s)**: #326

Fixes #335

** UI demo **:

https://github.com/hashgraph/hedera-smart-contracts/assets/66233296/7001f577-6e57-4d11-9da6-18f8bba3e0f0

**Checklist**

### DIFF
--- a/system-contract-dapp-playground/__tests__/hedera/erc20-interactions/index.test.ts
+++ b/system-contract-dapp-playground/__tests__/hedera/erc20-interactions/index.test.ts
@@ -1,0 +1,71 @@
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { getERC20TokenInformation } from '@/api/hedera/erc20-interactions';
+import { Contract } from 'ethers';
+
+describe('getERC20TokenInformation', () => {
+  const expectedName = 'TokenName';
+  const expectedSymbol = 'TKN';
+  const expectedTotalSupply = '1000000';
+  const expectedDecimals = '18';
+
+  // Mock baseContract object
+  const baseContract = {
+    name: jest.fn().mockResolvedValue(expectedName),
+    symbol: jest.fn().mockResolvedValue(expectedSymbol),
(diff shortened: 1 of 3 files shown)


system-contract-dapp-playground/__tests__/hedera/erc20-interactions/index.test.ts
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { getERC20TokenInformation } from '@/api/hedera/erc20-interactions';
```
Nana-EC, line 21: Interesting.
Q: Would import by relative file path not work and would this require you to compile to get any new changes?

system-contract-dapp-playground/src/components/contract-interaction/erc/erc-20/methods/TokenInformation.tsx
```diff
@@ -18,48 +18,195 @@
...
+  });
+
+  /**
+   * @dev handle execute token information queries
+   */
+  const handleExecutingMethods = async (method: 'name' | 'symbol' | 'totalSupply' | 'decimals') => {
+    let errTitle;
+
+    switch (method) {
+      case 'name':
```
Nana-EC, line 60: Move the shared logic between all the switch cases into a shared method.
Should be able to craft one that take in the token info label

Nana-EC (changes requested):
Looking good.
Some code reuse opportunities to reduce the amount of duplication

system-contract-dapp-playground/__tests__/hedera/erc20-interactions/index.test.ts
quiet-node: Hi Nana thanks for the question. Regards to the first part of the question, importing by relative file paths would work fine but I find using import aliases is cleaner. The import statement above when use relative paths would look sort of like this 
```
import { getERC20TokenInformation } from '../../../src/api/hedera/erc20-interactions';
```

For the second part of the question, if I understand your Q correctly then when new changes added it doesn't require to be recompiled. The import aliases are resolved at compile-time and used by the tsc to generate the correct import path in Javascript. After the code is done compiled, the import aliases are not relevant any more. It's a setting in the tsconfig file which can be found here https://github.com/hashgraph/hedera-smart-contracts/blob/main/system-contract-dapp-playground/tsconfig.json#L23

Nana-EC: Nice, thanks for the details 😃

Nana-EC (approved):
Nice improvement

PR #347, opened 2023-08-17
update: Update TokenCreateCustomContract to dynamically accept an off-chain keys array (#346)

### LINKED ISSUE #346: Update TokenCreateCustomContract to dynamically accept an off-chain keys array
### Problem

Currently, the TokenCreateCustomContract methods solely accommodate and establish a single public key for all key types (e.g. ADMIN, KYC, etc.). A more effective strategy would involve enhancing the methods to accept an array of keys typed IHederaTokenService.TokenKey, rather than a singular public key. This array can be preconfigured off-chain according to preferences and requirements.

### Solution

Replace the `bytes memory key` with `IHederaTokenService.TokenKey[] memory keys` inside TokenCreateCustomContract's methods

### Alternatives

_No response_

quiet-node:
**Description**: This PR 

- replaced the `bytes memory key` with `IHederaTokenService.TokenKey[] memory keys` inside TokenCreateCustomContract's methods so that the methods can accept a pre-configured keys array.
 
- modified unit tests to make sure the new logic works as expected

Fixes #346 

**Notes for reviewer**:

**Checklist**


test/hts-precompile/utils.js
```diff
@@ -668,6 +671,106 @@ class Utils {
...
+   *
+   * @notice IHederaTokenService.KeyValue supports a range of systems (e.g. ED25519, ECDSA secp256k1, etc.).
+   *         Refer to the documentation at https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L116
+   *         for more information on supported key types.
+   *
+   * @notice This function sets the `contractEVMAddress` (the address of the deployed HTS example contract) as the key for all key types (e.g. ADMIN, KYC, SUPPLY, etc.) for testing purposes.
+   *         If any other `keys` besides the `contractEVMAddress` are desired to be set, the keys must first sign an AccountUpdate transaction.
+   *         Refer to `this.updateAccountKeysViaHapi()` for usage examples on signing with `keys`.
+   */
+  static prepareTokenKey(contractEVMAddress, ECDSAPubKey) {
```
Nana-EC, line 686: So almost there but you'd want this method to be more customizeable.
Instead of this method returning an array of keys it should return a single key type.
Let callers contracts their own array.

test/hts-precompile/utils.js
```diff
@@ -668,6 +671,106 @@ class Utils {
...
+    return [
+      {
+        keyType: 1, // 0th bit: ADMIN
+        key: {
```
Nana-EC, line 690: This is conflicting logic.
The key type should be of one kind but you're setting all the value.
If the key is a contractId then set contract id, if it's ed25519 then set that, if it's an ecdsa then set that etc.

Nana-EC (changes requested):
Needs to provide a structure in utils for creating specific key types that callers can use to build the key structure they need

test/hts-precompile/utils.js
```diff
@@ -668,6 +671,158 @@ class Utils {
...
+   *                See https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L128
+   *                for more information
+   *
+   * @param keyValyeType INHERIT_ACCOUNT_KEY | CONTRACT_ID | ED25519 | SECP256K1 | DELEGETABLE_CONTRACT_ID
+   *
+   * @param key bytes value, public address of an account, or boolean
+   *            See https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L92
+   *                     for more information
+   */
+  static constructIHederaTokenKey(keyType, keyValueType, key) {
```
Nana-EC, line 688: Much better.
Nice improvement :) 
Should scrap `prepareTokenKeysArray()` and let callers combine this as needed

Nana-EC, line 688: nit:
```suggestion
  static constructIHederaTokenKey(keyType, keyValueType, value) {
```

test/hts-precompile/utils.js
```diff
@@ -668,6 +671,158 @@ class Utils {
...
+      (keyValueType !== 'INHERIT_ACCOUNT_KEY' &&
+        keyValueType !== 'CONTRACT_ID' &&
+        keyValueType !== 'ED25519' &&
+        keyValueType !== 'SECP256K1' &&
+        keyValueType !== 'DELEGETABLE_CONTRACT_ID')
+    ) {
+      return
+    }
+
+    return {
```
Nana-EC, line 707: Might be nice to create a const outside of this method that has the default values for all of this.
Then in a `switch` statement based on `keyValueType` you set the appropriate field.
This would simplify it and mean you can simplify the `if` above

Nana-EC (changes requested):

test/hts-precompile/utils.js
quiet-node: Fantastic idea! Updating the improvement soon!

test/hts-precompile/token-create/tokenCreateCustomContract.js
```diff
@@ -20,79 +20,108 @@
...
+      tokenCreateCustomContract.address
+    )
+
+    const supplyKey = utils.constructIHederaTokenKey(
+      'SUPPLY',
+      'CONTRACT_ID',
+      tokenCreateCustomContract.address
+    )
+
+    keys = [adminKey, pauseKey, kycKey, supplyKey]
```
Nana-EC, line 66: Q: So goal here is to have the main EOA be the signer of create, delete and pause transactions. Then have the contract be the authorizer of kyc, mint and burn transactions?
If so why and maybe you should add some comments to make this clear.
Notably wherever you but the EOA as the authorizer you will have to submit a direct call without going through the deployed contract.
It might be easier to start off with just the adminKey set and then in the PRs focused on those scenarios add the necessary key and tests etc

Nana-EC (changes requested):
Nice, almost there

Nana-EC (approved):
LG

PR #354, opened 2023-08-21
feat: Added API and unit tests to facilitate fungible and non-fungible token creations for TokenCreateCustom contract

quiet-node:
**Description**:
This PR adds a set of APIs and unit tests to support facilitating interacting with TokenCreateCustom contract. The methods exposed in this PR are

- createFungibleTokenPublic()
- createFungibleTokenWithCustomFeesPublic()
- createNonFungibleTokenPublic()
- createNonFungibleTokenWithCustomFeesPublic()

In addition, this PR also adds a couple of quick helper functions constructIHederaTokenKey() and prepareHederaTokenKeyArray() to help constructing an array of keys to assign to the token during creation

**Related issue(s)**: #314

Fixes partial #316

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
+++ b/system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
@@ -1,0 +1,139 @@
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+import { prepareHederaTokenKeyArray } from '@/utils/contract-interactions/HTS/helpers';
+import { Contract, isAddress } from 'ethers';
+
+/**
+ * @dev creates a Hedera fungible token
+ *
+ * @dev integrates tokenCreateCustomContract.createFungibleTokenPublic() and tokenCreateCustomContract.createFungibleTokenWithCustomFeesPublic()
+ *
+ * @param baseContract: ethers.Contract
+ *
+ * @param name: string
+ *
+ * @param symbol: string
+ *
+ * @param memo: string
+ *
+ * @param initialTotalSupply: number
+ *
+ * @param maxSupply: number
+ *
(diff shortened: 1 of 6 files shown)


system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
```diff
@@ -0,0 +1,237 @@
...
+        feeTokenAddress,
+        name,
+        symbol,
+        memo,
+        initialTotalSupply,
+        maxSupply,
+        decimals,
+        keyRes.hederaTokenKeys,
+        {
+          value: '35000000000000000000',
```
Nana-EC, line 102: Q: how did we come about the value and gasLimit values from?

system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
```diff
@@ -0,0 +1,237 @@
...
+          gasLimit: 1_000_000,
```
Nana-EC, line 103: Q: should these be parameter(s) for easy configurability?

Nana-EC (commented):
Looking good.
2 questions

system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
quiet-node: Oh great catch I got this 35 hbar number after a few tests for quite some time ago and never remember to got back to it. And yes you're right these should be parameters so can be passed in dynamically by the users. The fact is I didn't know we have a whole fee calculator on the Hedera website, so people can use the calculator to determine what the fee they should pass in. But yeah pushing the fix soon

quiet-node: Hmm on a second thought, it seems like a solid idea to have 'msgValue' set as a dynamic parameter. As for the 1_000_000 gas limit, I'm inclined to keep it as is for a smoother UX. Regarding 'msgValue' on the front-end, we can provide a link to the calculator, allowing users to determine the fee. However, adjusting the gas limit might confuse users, and I remember couple of time my transaction got rejected if the gas limits is improperly set. Any unused gas eventually gets returned to users, so there's no real loss on their end. Let me know what you think

Nana-EC: Yeah make `msg.value` an input
gasLimit refund max is 20% of what you put in so it's not all refunded.
The gasLimit changes per operation I would set it closer to what's needed.
DOn't need to hold up on the 1 million has limit but maybe open a ticket to revisit after phase 3 to have more realistic numbers or remove it so that `eth_estimateGas` can be used once HIP 584 is all done.

Nana-EC (approved):

system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
quiet-node: Yessir! Thanks for the input!

PR #355, opened 2023-08-22
feat: Added API and unit tests to facilitate token mint, token associating, and granting kyc for TokenCreateCustom contract

quiet-node:
**Description**:
This PR adds a set of APIs and unit tests to support facilitating interacting with TokenCreateCustom contract. The methods exposed in this PR are

- mintTokenPublic()
- mintTokenToAddressPublic()
- mintNonFungibleTokenToAddressPublic()
- associateTokenPublic
- associateTokensPublic()
- grantTokenKycPublic()

**Related issue(s)**: #314

Fixes partial #316

**Checklist**

### DIFF
--- a/system-contract-dapp-playground/__tests__/hedera/hts-interactions/index.test.ts
+++ b/system-contract-dapp-playground/__tests__/hedera/hts-interactions/index.test.ts
@@ -38,6 +38,10 @@ describe('createHederaFungibleToken test suite', () => {
   const grantingKYCAccount = '0x34810E139b451e0a4c67d5743E956Ac8990842A8';
   const txHash = '0x63424020a69bf46a0669f46dd66addba741b9c02d37fab1686428f5209bc759d';
   const returnedTokenAddress = '0x00000000000000000000000000000000000000000000000000000000000084b7';
+  const returnedMintedTokenEvent =
+    '000000000000000000000000000000000000000000000000000000000000000900000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000000';
+  const returnedTransferTokenEvent =
+    '0x000000000000000000000000000000000000000000000000000000000000abc700000000000000000000000034810e139b451e0a4c67d5743e956ac8990842a80000000000000000000000000000000000000000000000000000000000000001';
   const tokenName = 'WrappedHbar';
   const tokenSymbol = 'WHBAR';
   const tokenMemo = 'Wrapped Hbar';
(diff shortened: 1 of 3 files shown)


system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
```diff
@@ -239,8 +239,236 @@ export const createHederaNonFungibleToken = async (
...
+ *
+ * @return Promise<TokenCreateCustomSmartContractResult>
+ */
+export const mintHederaTokenToAddress = async (
+  baseContract: Contract,
+  tokenType: 'FUNGIBLE' | 'NON_FUNGIBLE',
+  hederaTokenAddress: string,
+  recipientAddress: string,
+  amountToMint: number,
+  metadata: string
```
Nana-EC, line 326: Q: shouldn't this be collection of metadata to support minting more than 1 NFT?

system-contract-dapp-playground/__tests__/hedera/hts-interactions/index.test.ts
```diff
@@ -94,6 +101,36 @@ describe('createHederaFungibleToken test suite', () => {
         hash: txHash,
       }),
     }),
+    mintTokenPublic: jest.fn().mockResolvedValue({
```
Nana-EC, line 104: With the `txHash` being the same for all maybe you should add an additional field that confirm the type of transaction.
Maybe similar to the transaction above there should be an event emitted you confirm

Nana-EC (changes requested):
Nice setup and progress.
1 question 1 suggestion

system-contract-dapp-playground/__tests__/hedera/hts-interactions/index.test.ts
quiet-node: Agree! I have put in the event filter for the token mint methods. For association and KYC, as the methods emit only ResponseCode events, and the APIs don't utilize these events so I thought I might have leave these out

Nana-EC (dismissed):

system-contract-dapp-playground/src/api/hedera/tokenCreateCustom-interactions/index.ts
```diff
@@ -239,8 +239,258 @@ export const createHederaNonFungibleToken = async (
...
+ *
+ * @param hederaTokenAddress: string
+ *
+ * @param amountToMint: number
+ *
+ * @param metadata: string[]
+ *
+ * @return Promise<TokenCreateCustomSmartContractResult>
+ */
+export const mintHederaToken = async (
```
Nana-EC, line 265: nit: I feel like these should all be in there own files and brought together in index.ts but that can be done later

Nana-EC (approved):

PR #358, opened 2023-08-25
feat: incorporated UI component for the NonFungibleTokenCreate method…

quiet-node:
**Description**:
This PR added the NonFungibleTokenCreate UI component which enables the use of both createNonFungibleTokenPublic() and createNonFungibleTokenWithFee() functions

**Related issue(s)**: #314 

Fixes partial #316 

** UI demo **

https://github.com/hashgraph/hedera-smart-contracts/assets/66233296/4c743426-b24a-494b-9ecb-df653869023a

**Checklist**


Nana-EC (approved):
Nice Lg.
Only item to consider is given the shared concepts between NFT and FT creation could we have shared more components and resumed them.
Some things to note for future improvements after Phase 3

PR #379, opened 2023-09-07
dapp-feat: finished integrating TokenTransfer API Methods (#319)

quiet-node:
**Description**:
This PR adds a suite of new interfaces, APIs for the TokenTransferContract, and unit tests for the APIs

**Related issue(s)**: #314 
 
Fixes partial #319 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/system-contract-dapp-playground/src/api/hedera/tokenTransfer-interactions/index.ts
+++ b/system-contract-dapp-playground/src/api/hedera/tokenTransfer-interactions/index.ts
@@ -48,15 +48,7 @@ export const transferCrypto = async (
   try {
     const tx = await baseContract.cryptoTransferPublic(transferList, tokenTransferList);
 
-    const txReceipt = await tx.wait();
-
-    // retrieve responseCode from event
-    const { data } = txReceipt.logs.filter(
-      (event: any) => event.fragment.name === 'ResponseCode'
-    )[0];
-
-    // @notice: 22 represents the predefined response code from the Hedera system contracts, indicating a successful transaction.
-    return { result: Number(data) === 22, transactionHash: txReceipt.hash };
+    return await handleContractResponse(tx);
   } catch (err: any) {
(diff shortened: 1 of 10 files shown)


system-contract-dapp-playground/src/types/contract-interactions/HTS/index.d.ts
```diff
@@ -72,6 +72,13 @@ interface TokenQuerySmartContractResult {
   err?: any;
 }
 
+/** @dev an interface for the results returned back from interacting with Hedera TokenTransfer smart contract */
+interface TokenTransferSmartContractResult {
```
Nana-EC, line 76: This look familiar. You sure we can't have a shared `SmartContractExecutionResult` and only add new interfaces when the properties differ?

system-contract-dapp-playground/src/api/hedera/tokenTransfer-interactions/index.ts
```diff
@@ -0,0 +1,330 @@
...
+): Promise<TokenTransferSmartContractResult> => {
+  // sanitize params
+  let sanitizeErr;
+  if (!isAddress(hederaTokenAddress)) {
+    sanitizeErr = 'Invalid token address';
+  } else if (!isAddress(sender)) {
+    sanitizeErr = 'Invalid sender address';
+  } else if (!isAddress(receiver)) {
+    sanitizeErr = 'Invalid receiver address';
+  } else if (quantity < 0) {
```
Nana-EC, line 227: Q: should we add a max for safe guarding? Check the protobuf and see if there's a max we can utilize.

Nana-EC (commented):
Looking good.
SOme comments.
Might be worth considering initial max values for serial numbers and amounts to safe guard.

Nana-EC (approved):

Issue #399, opened 2023-09-14
[System Contract DApp] Future features


quiet-node:
### Problem

A list of future features that would be nice to have for the dapp.

### Solution

Add more components to fulfill the requirements in the task list

```[tasklist]
### Tasks
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/400
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/401
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/405
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/420
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/421
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/427
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/434
- [ ] https://github.com/hashgraph/hedera-smart-contracts/issues/658
```

PR #423, opened 2023-09-25
dapp-feat: added READONLY transactions to main activity page (#400)

### LINKED ISSUE #400: [System Contract DApp] READONLY transactions (queries) should also be included in the /activity page
### Problem

Currently, the /activity page only shows the transactions that go through consensus. Readonly queries also needs to be included to this list.

### Solution

- When writing READON queries to storage, match the `TransactionResult` type.
- Tweaked utils/common/helpers.ts/prepareTransactionList() function to add READONLY to the list
- In /activity's transaction list table, add another column `transaction type`, the value could be TRANSACTION or QUERY

### Alternatives

_No response_

quiet-node:
**Description**: This PRs reworked the caching readonly transactions for ERC20 and ERC721 methods. Now all request types both queries and transactions share the same ITransactionResult interface which make all the transactions more consistent in structure.

**Related issue(s)**: #399 

Fixes #400 

** UI Demo **:

https://github.com/hashgraph/hedera-smart-contracts/assets/66233296/538e683e-c41a-4aa9-afdd-0cfe11cbf848

**Notes for reviewer**:

**Checklist**


ebadiere (commented):
Can the recording include an audio explanation of what's going on?  The initial impression with no sound is not good.

quiet-node:
> Have we considered adding tests around this, using perhaps synpress?

Since we didn't have much time at the beginning so there are only tests for the APIs for now but not for the UI/UX for now.

Nana-EC (approved):

PR #437, opened 2023-10-06
Block info tests.

### LINKED ISSUE #430: [Solidity Support] Block Properties
### Problem

Occasionally items have shown up that surfaced inequivalence at the language layer.

### Solution

Add support for block properties

- Add a /contracts/solidity/block folder
- Cover the tasklist below

```[tasklist]
### Block property Tasks
- [x] block.basefee (uint): current block’s base fee (EIP-3198 and EIP-1559)
- [x] blockhash(uint blockNumber) returns (bytes32): hash of the given block when blocknumber is one of the 256 most recent blocks; otherwise returns zero
- [x] `block.coinbase` (`address payable`): current block miner’s address
- [x] `block.difficulty` (`uint`): current block difficulty (`EVM < Paris`). For other EVM versions it behaves as a deprecated alias for `block.prevrandao` ([[EIP-4399](https://eips.ethereum.org/EIPS/eip-4399)](https://eips.ethereum.org/EIPS/eip-4399) )
- [x] `block.gaslimit` (`uint`): current block gaslimit
- [x] `block.number` (`uint`): current block number
- [x] `block.prevrandao` (`uint`): random number provided by the beacon chain (`EVM >= Paris`)
- [x] `block.timestamp` (`uint`): current block timestamp as seconds since unix epoch
```

Ref: https://docs.soliditylang.org/en/latest/units-and-global-variables.html

ebadiere:
Adds a blockNumber test.  This is part of issue #430 

**Related issue(s)**:

Fixes #430


.github/workflows/tests.yml
```diff
@@ -17,6 +17,12 @@ permissions:
   pull-requests: write
 
 jobs:
+  BlockInfo:
```
Nana-EC, line 20: DOn't add these the normal Tests.
It's gonna end up being too large.
Please create a new workflow that can be kicked off manually for Solidity EVM equivalence coverage.
Can do this is a separate PR

contracts/hip-583/ContractTransferTx.sol
```diff
@@ -8,8 +8,6 @@ contract ContractTransferTx {
     function transferTo(address payable _to, uint256 _amount) public {
         require(address(this).balance > _amount, "Insufficient contract balance");
 
-        _to.call{value : _amount}("");
```
Nana-EC, line 11: Q: Why was this removed?

contracts/solidity/block/BlockInfo.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    function getBlockTimestamp() external view returns (uint256) {
+        return block.timestamp;
+    }
+
+    // should behave like prevrandao
+    function getBlockDifficulty() external view returns (uint256) {
+        return block.difficulty;
+    }
+}
```
Nana-EC, line 38: nit: new line for git happiness

local.env
```diff
@@ -8,6 +8,6 @@ OPERATOR_KEY_A= 302e020100300506032b65700422042091132178e72057a1d7528025956fe39b
 # ECDSA HEX Encoded Private Key
 HEX_PRIVATE_KEY_A= 0x2e1d968b041d84dd120a5860cee60cd83f9374ef527ca86996317ada3d0d03e7
 HEX_PRIVATE_KEY_B= 0x45a5a7108a18dd5013cf2d5857a28144beadc9c70b3bdbd914e38df4e804b8d8
-RETRY_DELAY=3000 # ms
-MAX_RETRY=20
+RETRY_DELAY=4000 # ms
```
Nana-EC, line 11: Q: why the increase? Are we seeing this to be necessary?

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+const { expect } = require('chai')
+const { ethers } = require('hardhat')
+const { BigNumber } = require('ethers');
+const Constants = require('../../constants')
+
+describe('BlockInfo Test Suite', function () {
```
Nana-EC, line 25: Let's add a tag for easy running identification. Maybe
```suggestion
describe('@solidityevmequiv BlockInfo Test Suite', function () {
```

Nana-EC (changes requested):
Nice progress, some questions and suggestions

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+  let blockInfo, provider, signers
+
+  before(async function () {
+    signers = await ethers.getSigners()
+    // const ethers = hre.ethers;
```
attest08, line 30: Can you remove the commented line? If you do not plan to use it, of course.

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+    blockInfo = await factory.  deploy()
+  })
+
+  // EIP-1559 does not apply to Hedera
+  it('should be able to execute getBlockBaseFee()', async function () {
+    const blockBaseFee = await blockInfo.getBlockBaseFee()
+    expect(blockBaseFee).to.equal(0)
+  })
+
+  xit('should be able to get the hash of a given block when the block number is one of the 256 most recent blocks', async function () {
```
attest08, line 43: Can you add comments with the reason why the tests are excluded.

contracts/solidity/block/BlockInfo.sol
ebadiere: Done.

test/solidity/block/BlockInfo.js
ebadiere: Done.

AlfredoG87 (dismissed):
LGTM


Just one question, why are we using Solidity version `8.7.0` why not `8.6.0` or `8.21.0`?

test/solidity/block/BlockInfo.js
ebadiere: Good catch.  Done

local.env
ebadiere: It took more than three runs to get the tests to pass, so I tweaked the values to see if it would reduce the number of runs to get all the tests to pass.

contracts/hip-583/ContractTransferTx.sol
Nana-EC: How can this be true? 
Without this line this method only does a require and then an event emission. 
If it's not being used then where is the transfer done?

Nana-EC (dismissed):
Unblocking.
If we can get to the bottom of why we need such a wide range of retry support that would be good.
If not let's open a ticket in follow up

Nana-EC (approved):

Issue #431, opened 2023-10-04
[Solidity Support] Transaction properties


Nana-EC:
### Problem

The Hedera network supports solidity as one of the smart contracts development languages.
Occasionally items have shown up that surfaced inequivalence at the language layer.

### Solution

Add support for transaction properties

- Add a /contracts/solidity/transaction folder
- Cover the tasklist below

```[tasklist]
### Transaction property Tasks
- [x] `gasleft() returns (uint256)`: remaining gas
- [x] `msg.data` (`bytes calldata`): complete calldata
- [x] `msg.sender` (`address`): sender of the message (current call)
- [x] `msg.sig` (`bytes4`): first four bytes of the calldata (i.e. function identifier)
- [x] `msg.value` (`uint`): number of wei sent with the message
- [x] `tx.gasprice` (`uint`): gas price of the transaction
- [x] `tx.origin` (`address`): sender of the transaction (full call chain)
```

 Ref: https://docs.soliditylang.org/en/latest/units-and-global-variables.html

PR #453, opened 2023-10-11
Adding tests for the Transactions in Solidity

stefan-stefanooov:
**Description**:
Add support for transaction properties

**Related issue(s)**:
https://github.com/hashgraph/hedera-smart-contracts/issues/431

Fixes #

**Notes for reviewer**:

### DIFF
--- a/contracts/solidity/transaction/Secondary.sol
+++ b/contracts/solidity/transaction/Secondary.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.20;
+
+contract Secondary {
+    function getTxOrigin() external view returns (address) {
+        return tx.origin;
+    }
+
+    function getMsgSender() external view returns (address) {
+        return msg.sender;
+    }
+}
--- a/contracts/solidity/transaction/Transaction.sol
+++ b/contracts/solidity/transaction/Transaction.sol
@@ -1,18 +1,18 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.20;
 
-import { Secondary } from "./Secondary.sol";
+import { MessageFrameAddresses } from "./MessageFrameAddresses.sol";
 
 contract Transaction {
     string public message;
     uint myInteger;
-    address secondaryAddr;
-    Secondary secondaryContract;
+    address messageFrameAddresses;
+    MessageFrameAddresses mfContract;
     event MsgValue(uint256);
 
-    constructor(address secondaryAddress) {
-       secondaryAddr = secondaryAddress;
-       secondaryContract = Secondary(payable(secondaryAddress));
+    constructor(address addr) {
+       messageFrameAddresses = addr;
+       mfContract = MessageFrameAddresses(payable(addr));
     }
 
     function checkGasleft() external view returns (uint256) {
@@ -43,10 +43,10 @@ contract Transaction {
     }
 
     function getTxOriginFromSecondary() external view returns (address) {
-        return secondaryContract.getTxOrigin();
+        return mfContract.getTxOrigin();
     }
 
     function getMsgSenderFromSecondary() external view returns (address) {
-        return secondaryContract.getMsgSender();
+        return mfContract.getMsgSender();
     }
 }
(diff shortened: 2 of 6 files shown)


contracts/solidity/transaction/Transaction.sol
```diff
@@ -0,0 +1,54 @@
...
+    function getMessageValue() public payable {
+        emit MsgValue(msg.value);
+    }
+
+    function getGasPrice() public view returns (uint256) {
+        return tx.gasprice;
+    }
+
+    function getTxOriginFromSecondary() public view returns (address) {
+        address secondaryOrigin = secondaryContract.getTxOrigin();
```
natanasow, line 46: The local `secondaryOrigin` variable is redundant. The method could be simplified to 
```solidity
function getTxOriginFromSecondary() public view returns (address) {
    return secondaryContract.getTxOrigin();
}
```

test/solidity/transaction/transaction.js
```diff
@@ -0,0 +1,118 @@
...
+    const signers = await ethers.getSigners()
+    wallet = signers[0];
+    senderWalletAddr = await wallet.getAddress()
+  })
+
+  it('gasleft() returns (uint256): remaining gas', async function () {
+    const gasLeft = await contractTr.checkGasleft()
+    
+    expect(gasLeft).to.exist
+    expect(ethers.BigNumber.isBigNumber(gasLeft)).to.be.true
```
Nana-EC, line 53: Would  be good here to pass in a gas amount and check that the `0 < gasLeft < gasSent`

test/solidity/transaction/transaction.js
```diff
@@ -0,0 +1,118 @@
...
+    
+    // to compare with the value sent, we need to convert to tinybar
+    expect(amount.mul('10000000000')).to.equal(valueToSend)
+  })
+
+  it('tx.gasprice (uint): gas price of the transaction', async function () {
+    const gasPrice = await contractTr.getGasPrice()
+
+    expect(gasPrice).to.exist
+    expect(ethers.BigNumber.isBigNumber(gasPrice)).to.be.true
```
Nana-EC, line 106: Add greater than 0 check here and any other numbers.
0 gets used often as a default and can signal a bug

contracts/solidity/transaction/Secondary.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.20;
+
+contract Secondary {
```
Nana-EC, line 4: nit: Looking for less general names. What about `MessageFrameAddresses` or `MessageFrameAccounts`?
Ideas?

Nana-EC (changes requested):
nice, some suggestions

test/solidity/transaction/transaction.js
stefan-stefanooov: Thanks, good to know.
I added `greater than 0` check. 
I am not sure I understand the second part of the comment `...and any other numbers.`

contracts/solidity/transaction/Secondary.sol
stefan-stefanooov: `MessageFrameAddresses` sounds good to me.
Thanks ... got a bit lazy with the naming I guess.

test/solidity/transaction/transaction.js
stefan-stefanooov: Done.

test/solidity/transaction/transaction.js
Nana-EC: Yeah i just meant any other values you may be checking in EVM responses you might want to confirm non zero also

Nana-EC (approved):
LGTM

test/solidity/transaction/transaction.js
```diff
@@ -0,0 +1,119 @@
...
+ * limitations under the License.
+ *
+ */
+
+const { expect } = require('chai')
+const { ethers } = require('hardhat')
+const Constants = require('../../constants')
+const Utils = require('../../hts-precompile/utils')
+
+describe('PrngSystemContract tests', function () {
```
quiet-node, line 26: Shouldn't this be 'transaction tests' instead?

Nana-EC: @quiet-node please fix this and I can sign off

Nana-EC (approved):
LGTM

stefan-stefanooov:
Thanks for the help here @quiet-node ;)

PR #455, opened 2023-10-11
Adding solidity workflow.

ebadiere:
Adding new solidity tests GH test workflow.  Related issues: #430 

Fixes #

### DIFF
--- a/.github/workflows/solidity-tests.yml
+++ b/.github/workflows/solidity-tests.yml
@@ -13,7 +13,7 @@ permissions:
   pull-requests: write
 
 jobs:
-  BlockInfo:
+  SolidityEquivalence:
     name: BlockInfo Test Suite
     uses: ./.github/workflows/test-workflow.yml
     with:
--- a/.github/workflows/solidity-tests.yml
+++ b/.github/workflows/solidity-tests.yml
@@ -1,4 +1,4 @@
-name: Solidity Tests CI
+name: Solidity Equivalence Tests CI
 
 on:
   workflow_dispatch:
@@ -17,7 +17,7 @@ jobs:
     name: BlockInfo Test Suite
     uses: ./.github/workflows/test-workflow.yml
     with:
-      testfilter: BlockInfo
+      testfilter: @solidityequiv
 
 
   PublishResults:
--- a/.github/workflows/solidity-tests.yml
+++ b/.github/workflows/solidity-tests.yml
@@ -1,11 +1,7 @@
 name: Solidity Tests CI
 
 on:
-  pull_request:
-    branches: [main, release/**]
-  push:
-    branches: [main, release/**]
-    tags: [v*]
+  workflow_dispatch:
 
 concurrency:
   group: tests-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/solidity-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Solidity Tests CI
+
+on:
```
Nana-EC, line 3: Make this manual. There will be a lot of tests so we don't want to run them always so it should be from a workflow kick off

.github/workflows/solidity-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+  group: tests-${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: write
+  checks: write
+  pull-requests: write
+
+jobs:
+  BlockInfo:
```
Nana-EC, line 20:
```suggestion
  SolidityEquivalence:
```

.github/workflows/solidity-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+    name: BlockInfo Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: BlockInfo
```
Nana-EC, line 24: Let's use a reusable tag
```suggestion
      testfilter: @solidityequiv
```

.github/workflows/solidity-tests.yml
```diff
@@ -0,0 +1,45 @@
+name: Solidity Tests CI
```
Nana-EC, line 1: nit:
```suggestion
name: Solidity Equivalence Tests CI
```

Nana-EC (changes requested):
Nice.
Some suggestions to make it more general

.github/workflows/solidity-tests.yml
ebadiere: Done.

.github/workflows/solidity-tests.yml
ebadiere: By doing this I assume we are going to run all of the solidity equivalence tests in one job.  Is that what we want?  I am not sure that's scalable.

.github/workflows/solidity-tests.yml
ebadiere: Done, but I still question putting all tests in one umbrella job.

.github/workflows/solidity-tests.yml
ebadiere: Done.

.github/workflows/solidity-tests.yml
Nana-EC: You can break them out but it shouldn't be by file.
We can group them in larger buckets.
What would you suggest?

ebadiere: I guess the term "SolidityEquivalence" is what I am questioning here.  I think it makes sense to start with this, so I'll make the update, but the term seems to be an umbrella term for this whole effort which will result in many more tests.

ebadiere: Done.

Nana-EC (approved):

PR #486, opened 2023-10-17
Batch Solidity Equivalence Tests

Nana-EC:
**Description**:

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/test-workflow.yml
+++ b/.github/workflows/test-workflow.yml
@@ -40,11 +40,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Install Foundry
-        uses: foundry-rs/foundry-toolchain@v1
-        with:
-          version: nightly
-
       - name: Start the local node
         run: npx hedera start -d --network local
         timeout-minutes: 5
--- a/.github/workflows/solidity-tests.yml
+++ b/.github/workflows/solidity-tests.yml
@@ -1,7 +1,12 @@
-name: Solidity Equivalence Tests CI
+name: Solidity Equivalence Tests
 
 on:
-  workflow_call:
+  workflow_dispatch:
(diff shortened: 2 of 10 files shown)


.github/workflows/test-workflow.yml
```diff
@@ -40,11 +40,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Install Foundry
```
Nana-EC, line 43: @mshakeg any idea why the Foundry install is still needed in the `test-workflow.yml` for the non foundry executions?
I realize we have to leave it in but curious if we know why it causes the `HardhatFoundryError: Couldn't run `forge`. Please check that your foundry installation is correct` error?

Will restore it but let me know if you have context here.

mshakeg: The project now incorporates the [hardhat-foundry plugin](https://hardhat.org/hardhat-runner/plugins/nomicfoundation-hardhat-foundry). This plugin enables Hardhat to utilize dependencies from the `./lib` folder, which are installed using `forge`. Consequently, the plugin tries executing `forge install` to make these dependencies accessible to Hardhat. While the Foundry configuration permits Foundry to utilize dependencies from `./node_modules`, if you think that the contracts won't rely on the `forge`-installed dependencies as is currently the case, you can consider removing the `hardhat-foundry` plugin.

Nana-EC: I'm sure they will more over time as your PR showed.
Also looking at the OZ library they eventually reached a point of balance with hardhat and foundry support for the same contract so it's not inconceivable in the future we could also.
It's fine as is just wanted to understand the connection since the workflow doesn't run foundry tests explicitly.
Thanks

Issue #471, opened 2023-10-17
[Solidity Support] Defaults values


Nana-EC:
### Problem

The Hedera network supports solidity as one of the smart contracts development languages.
Occasionally items have shown up that surfaced in-equivalence at the language layer.

### Solution

Add examples for assembly address operations

- Add `/contracts/solidity/defaults`
- Cover the tasklist below to read default values

```[tasklist]
### Defaults Tasks
- [x] `uint<M>` cover different variations
- [x] `int<M>` cover different variations
- [ ] `fixed<M>x<N>`  cover different variations
- [ ] `unfixed<M>x<N>`  cover different variations
- [x] `bytes<M>`
- [x] `string`
- [x] `<type>[]`
- [x] `address`
- [x] `mapping`
```

Declare variable and read to verify default values

stefan-stefanooov:
As far as I understand the "fixed" type is still not fully supported in Solidity.

PR #497, opened 2023-10-18
Adding tests for the default values of all types in Solidity

stefan-stefanooov:
**Description**:
Adding tests for the default values of all types in Solidity

**Related issue(s)**:
https://github.com/hashgraph/hedera-smart-contracts/issues/471

### DIFF
--- a/contracts/solidity/defaults/Defaults.sol
+++ b/contracts/solidity/defaults/Defaults.sol
@@ -59,6 +59,10 @@ contract Defaults {
         bytes[] bytesArr;
     }
 
+    mapping(string => uint) public strUintMap;
+    mapping(address => bool) public addrBoolMap;
+    mapping(int => bytes) public bytesBytesMap;
+
     function getUintDefaults() external pure returns (UintDefaults memory) {
         UintDefaults memory defaults;
         return defaults;
--- a/contracts/solidity/defaults/Defaults.sol
+++ b/contracts/solidity/defaults/Defaults.sol
@@ -1,0 +1,104 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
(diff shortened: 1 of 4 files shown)


contracts/solidity/defaults/Defaults.sol
```diff
@@ -0,0 +1,108 @@
...
+        UintDefaults memory defaults;
+        return defaults;
+    }
+
+    function getIntDefaults() external pure returns (UintDefaults memory) {
+        UintDefaults memory defaults;
+        return defaults;
+    }
+
+    // Not supported by solidity yet
```
Nana-EC, line 76: Q: Interesting, any docs to add a note here?

test/solidity/defaults/defaults.js
```diff
@@ -0,0 +1,85 @@
+const { expect } = require('chai')
+const { ethers } = require('hardhat')
+const Constants = require('../../constants')
+
+describe('Solidity Defaults', function () {
```
Nana-EC, line 5: Add test tag

Nana-EC (commented):
Looking good.
Please add tag and ref for the not supported aspect

contracts/solidity/defaults/Defaults.sol
stefan-stefanooov: You can find the documentation: [here](https://docs.soliditylang.org/en/latest/types.html#fixed-point-numbers)

image

Nana-EC: Yes, I meant you could add an in code comment so that it's not perceived as Hedera doesn't support it and rather the language itself.

stefan-stefanooov: Done

PR #500, opened 2023-10-18
Block information tests.

### LINKED ISSUE #430: [Solidity Support] Block Properties

ebadiere:
Tests covering the Block properties.

**Related issue(s)**: 

Fixes #430 

### DIFF
--- a/test/solidity/block/BlockInfo.js
+++ b/test/solidity/block/BlockInfo.js
@@ -22,7 +22,7 @@ const { ethers } = require('hardhat')
 const { BigNumber } = require('ethers');
 const Constants = require('../../constants')
 
-describe('@solidityevmequiv BlockInfo Test Suite', function () {
+describe('@solidityequiv1 BlockInfo Test Suite', function () {
   let blockInfo, provider, signers
 
   before(async function () {
--- a/test/solidity/block/BlockInfo.js
+++ b/test/solidity/block/BlockInfo.js
@@ -43,8 +43,14 @@ describe('@solidityequiv1 BlockInfo Test Suite', function () {
   it('should be able to get the hash of a given block when the block number is one of the 256 most recent blocks', async function () {
     const blockNumber = await provider.getBlockNumber()
     const block = await provider.getBlock(blockNumber)
-    const blockHash = await blockInfo.getBlockHash()
-    console.log(`Block hash: ${blockHash}`)
+    try {
+      const blockHash = await blockInfo.getBlockHash()
+    } catch(e) {
+      expect(e.code).to.equal('CALL_EXCEPTION')
+      expect(e.message).to.contain('missing revert data in call exception')
+      expect(e.reason).to.contain('missing revert data in call exception; Transaction reverted without a reason string')      
+    }
+    
   })
   
   it('should get the current block coinbase which is the hedera network account', async function () { 
@@ -54,15 +60,30 @@ describe('@solidityequiv1 BlockInfo Test Suite', function () {
   })
 
   // Turn off until mirror node issue is resolved: https://github.com/hashgraph/hedera-mirror-node/issues/7036
-  xit('should get the current block prevrandao using block.prevrandao', async function () { 
-    const prevrandao = await blockInfo.getBlockPrevrando()  
-    expect(BigNumber.isBigNumber(prevrandao)).to.be.true
+  it('should get the current block prevrandao using block.prevrandao', async function () { 
+    try{
+      const prevrandao = await blockInfo.getBlockPrevrando()  
+    } catch (e) {
+      expect(e.code).to.equal('CALL_EXCEPTION')
+      expect(e.message).to.contain('missing revert data in call exception')
+      expect(e.reason).to.contain('missing revert data in call exception; Transaction reverted without a reason string')
+    }
+    // Uncomment when mirror node issue is resolved
+    // expect(BigNumber.isBigNumber(prevrandao)).to.be.true
   })
 
   // Turn off until mirror node issue is resolved: https://github.com/hashgraph/hedera-mirror-node/issues/7036
-  xit('should get the current block difficulty using block.difficulty (replaced by prevrandao)', async function () { 
-    const difficulty = await blockInfo.getBlockDifficulty()  
-    expect(BigNumber.isBigNumber(difficulty)).to.be.true
+  it('should get the current block difficulty using block.difficulty (replaced by prevrandao)', async function () { 
+    try {
+      const difficulty = await blockInfo.getBlockDifficulty()
+    } catch (e) {
+      expect(e.code).to.equal('CALL_EXCEPTION')
+      expect(e.message).to.contain('missing revert data in call exception')
+      expect(e.reason).to.contain('missing revert data in call exception; Transaction reverted without a reason string')
+
+    }
+    // Uncomment when mirror node issue is resolved      
+    // expect(BigNumber.isBigNumber(difficulty)).to.be.true
   })
 
   it('should get the block gas limit', async function () { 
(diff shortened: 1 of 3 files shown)


test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+
+  before(async function () {
+    signers = await ethers.getSigners()
+    provider = ethers.getDefaultProvider();
+    
+    const factory = await ethers.getContractFactory(Constants.Path.BLOCK_INFO)
+    blockInfo = await factory. deploy({ gasLimit: 15000000 })
+  })
+
+  // EIP-1559 does not apply to Hedera
```
Nana-EC, line 36: Can you clarify intent here.
Hedera explicitly supports EIP 1559 transaction types.

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+  it('should be able to execute getBlockBaseFee()', async function () {
+    const blockBaseFee = await blockInfo.getBlockBaseFee()
+    expect(blockBaseFee).to.equal(0)
+  })
+
+  // Turn off until mirror node issue is resolved: https://github.com/hashgraph/hedera-mirror-node/issues/7045
+  xit('should be able to get the hash of a given block when the block number is one of the 256 most recent blocks', async function () {
```
Nana-EC, line 43: Please turn test on. Instead of turning test off expect the incorrect value but you can leave the ticket issue here. Once that issue is fixed we should see this test start to fail.
This avoids where we forget to come back to fix it.

Same case for other tests in this case

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+
+  const date = new Date(value * 1000);
+  if (isNaN(date)) {
+      return false;
+  }
+
+  const year = date.getUTCFullYear();
+  return year >= 1970 && year <= 2030;
+}
+
```
Nana-EC, line 100: nit: remove all the extra lines.

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,105 @@
...
+      return false;
+  }
+
+  const date = new Date(value * 1000);
+  if (isNaN(date)) {
+      return false;
+  }
+
+  const year = date.getUTCFullYear();
+  return year >= 1970 && year <= 2030;
```
Nana-EC, line 98: Q: why the 2030 max choice?

Nana-EC (changes requested):
Update tests to expect current node behaviour and other nits

test/solidity/block/BlockInfo.js
ebadiere: The test already fails.  The following error is returned : 
```
Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"from":"0x05FbA803Be258049A27B820088bab1cAD2058871","gasLimit":{"type":"BigNumber","hex":"0x89a758"},"to":"0xA4096Ee370402B804288a967dE244844287028AD","data":"0x9663f88f","accessList":null}, error={"name":"ProviderError","_stack":"ProviderError: [Request ID: 4218bfa1-39a1-45cb-9313-79fc59bd9ad2] execution reverted: Expected 32 bytes but got 48\n    at HttpProvider.request (/Users/user/projects/hedera-smart-contracts/node_modules/hardhat/src/internal/core/providers/http.ts:88:21)\n    at processTicksAndRejections (node:internal/process/task_queues:95:5)\n    at EthersProviderWrapper.send (/Users/user/projects/hedera-smart-contracts/node_modules/@nomiclabs/hardhat-ethers/src/internal/ethers-provider-wrapper.ts:13:20)","code":-32008,"_isProviderError":true,"data":""}, code=CALL_EXCEPTION, version=providers/5.7.2)
```

test/solidity/block/BlockInfo.js
ebadiere: Done.

test/solidity/block/BlockInfo.js
ebadiere: Removed the max year.

test/solidity/block/BlockInfo.js
ebadiere: Turned on for now.

test/solidity/block/BlockInfo.js
Nana-EC: Would be better to just say base fee don't adjust per block. The EIP itself does apply

test/solidity/block/BlockInfo.js
Nana-EC: Blank lines are still there

test/solidity/block/BlockInfo.js
```diff
@@ -0,0 +1,104 @@
...
+    console.log(`Block hash: ${blockHash}`)
+  })
+  
+  it('should get the current block coinbase which is the hedera network account', async function () { 
+    const coinbase = await blockInfo.getMinerAddress()
+    // 0.0.98 is the Hedera network account.  Alias is 0x0000000000000000000000000000000000000062
+    expect(coinbase).to.equal('0x0000000000000000000000000000000000000062')
+  })
+
+  // Turn off until mirror node issue is resolved: https://github.com/hashgraph/hedera-mirror-node/issues/7036
```
Nana-EC, line 56: Please turn on the test.
Catch the current failure and in comments note the bug and that once it resolved no error should occur.
We basically want to ensure that the moment its fixed on mirror node these tests will break and we'll have to fix it. Otherwise it'd be easy for these tests to remain dormant

Nana-EC (changes requested):
2 items remaining

test/solidity/block/BlockInfo.js
ebadiere: Done.  Sorry for the delay.

Nana-EC: @ebadiere thanks.
It's missing a check to make sure it catches an error.
If the error isn't thrown the test will pass without us realizing

test/solidity/block/BlockInfo.js
Nana-EC: @ebadiere still there

Nana-EC (changes requested):
A few outstanding changes

test/solidity/block/BlockInfo.js
ebadiere: Done.

test/solidity/block/BlockInfo.js
ebadiere: Handles expected call exception.

test/solidity/block/BlockInfo.js
ebadiere: Removed sorry.

test/solidity/block/BlockInfo.js
ebadiere: Done.

Nana-EC (approved):

Issue #482, opened 2023-10-17
[Solidity Support] Non existing accounts support


Nana-EC:
### Problem

The Hedera network supports solidity as one of the smart contracts development languages.
Occasionally items have shown up that surfaced in-equivalence at the language layer.

### Solution

Add examples for non existing account operations

- Add `/contracts/solidity/account/NonExisting.sol`
- Cover the task list below

```[tasklist]
### Account Tasks
- [x] Use `call` on a non existing account
- [x] Use `delegatecall ` on a non existing
- [x] Use `staticcall` on a non existing account
```

According to EVM and solidity the first return should be true.
However, we're currently expecting failures of the Hedera INVALID error types, so these tests should test for the invalid error until the issue is resolved on services. In this was once resoled these tests should fail and require changing.

PR #501, opened 2023-10-18
Adding tests for NonExistant account

stefan-stefanooov:
**Description**:
Adding tests for NonExistant account

**Related issue(s)**:
https://github.com/hashgraph/hedera-smart-contracts/issues/482

### DIFF
--- a/contracts/solidity/account/NonExisting.sol
+++ b/contracts/solidity/account/NonExisting.sol
@@ -1,7 +1,16 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.20;
 
+import { NonExtDup } from "./NonExtDup.sol";
+
 contract NonExisting {
+    NonExtDup duplicate;
+
+    constructor(address addr) {
+        // solhint-disable-previous-line no-empty-blocks
+        duplicate = NonExtDup(addr);
+    }
+
     function callOnNonExistingAccount(address nonExistingAddr) external {
         nonExistingAddr.call(
             abi.encodeWithSignature("doesNotExist()")
@@ -18,5 +27,9 @@ contract NonExisting {
         nonExistingAddr.staticcall(
             abi.encodeWithSignature("doesNotExist()")
         );
+    }
+
+    function balanceNoneExistingAddr(address nonExistingAddr) external view returns (uint256) {
+        return nonExistingAddr.balance;
     }
 }
--- a/contracts/solidity/account/NonExisting.sol
+++ b/contracts/solidity/account/NonExisting.sol
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+contract NonExisting {
+    function callOnNonExistingAccount(address nonExistingAddr) external {
+        nonExistingAddr.call(
+            abi.encodeWithSignature("doesNotExist()")
+        );
+    }
+
+    function delegatecallOnNoneExistingAccount(address nonExistingAddr) external {
+        nonExistingAddr.delegatecall(
+            abi.encodeWithSignature("doesNotExist()")
+        );
+    }
+
+    function staticcallOnNoneExistingAccount(address nonExistingAddr) external {
+        nonExistingAddr.staticcall(
+            abi.encodeWithSignature("doesNotExist()")
+        );
+    }
+}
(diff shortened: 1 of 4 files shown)


contracts/solidity/account/NonExisting.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+contract NonExisting {
+    function callOnNonExistingAccount(address nonExistingAddr) external {
+        nonExistingAddr.call{value: 0}(
```
natanasow, line 6: if the value is 0, we can remove it IMO

contracts/solidity/account/NonExisting.sol
```diff
@@ -0,0 +1,22 @@
...
+            abi.encodeWithSignature("doesNotExist()")
+        );
+    }
+
+    function delegatecallOnNoneExistingAccount(address nonExistingAddr) external {
+        (bool success, bytes memory data) = nonExistingAddr.delegatecall(
```
natanasow, line 12: we can remove `(bool success, bytes memory data)` if it's not used in the function

contracts/solidity/account/NonExisting.sol
```diff
@@ -0,0 +1,22 @@
...
+            abi.encodeWithSignature("doesNotExist()")
+        );
+    }
+
+    function staticcallOnNoneExistingAccount(address nonExistingAddr) external {
+        (bool success, bytes memory data) = nonExistingAddr.staticcall(
```
natanasow, line 18: same

test/solidity/account/nonExisting.js
```diff
@@ -0,0 +1,44 @@
+const { expect, assert } = require('chai')
+const { ethers } = require('hardhat')
+const Constants = require('../../constants')
+
+describe('Solidity Account non Existing', function () {
+  let contract, wallet
+
+  before(async function () {
+    const signers = await ethers.getSigners()
+    wallet = signers[0]
```
natanasow, line 10: we should create and use one random address instead of shifting the current wallet address in each test
```typescript
const randomAddress = ethers.Wallet.createRandom().address;
```

test/solidity/account/nonExisting.js
```diff
@@ -0,0 +1,44 @@
...
+    const factory = await ethers.getContractFactory(Constants.Contract.NonExisting)
+    contract = await factory.deploy()
+  })
+
+  it('should confirm `call` on a non existing account', async function () {
+    try {
+        const addr = wallet.address << 3;
```
natanasow, line 17: can be removed

test/solidity/account/nonExisting.js
```diff
@@ -0,0 +1,44 @@
...
+        const tx = await contract.callOnNonExistingAccount(addr)
+        await tx.wait()
+    } catch (err) {
+        expect(err.reason).to.equal("invalid address or ENS name")
+    }
+  })
+
+  it('should confirm `delegatecall` on a non existing account', async function () {
+    try {
+        const addr = wallet.address << 3;
```
natanasow, line 27: same

natanasow (changes requested):
LGTM, some nits

contracts/solidity/account/NonExisting.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+contract NonExisting {
```
Nana-EC, line 4: None of the methods use the address param being passed in

test/solidity/account/nonExisting.js
```diff
@@ -0,0 +1,44 @@
+const { expect, assert } = require('chai')
```
Nana-EC, line 1: Missing license header

Nana-EC (changes requested):

Nana-EC (commented):
LG
Currently the tests use a non existing address as the base for the call.
What about when the calls is to a non existing address e.g. a contract creation for an address that doesn't exist or a function call (code size or balance) on an address that doesn't exist. The last example might already be covered in other recent solidity code so double check the epic or current tests.

stefan-stefanooov:
> LG Currently the tests use a non existing address as the base for the call. What about when the calls is to a non existing address e.g. a contract creation for an address that doesn't exist or a function call (code size or balance) on an address that doesn't exist. The last example might already be covered in other recent solidity code so double check the epic or current tests.

Addressed in the new version :)

Nana-EC (approved):

natanasow (approved):
LGTM

Issue #489, opened 2023-10-18
[Solidity Support]  Simple auction example


Nana-EC:
### Problem

The Hedera network supports solidity as one of the smart contracts development languages.
Occasionally items have shown up that surfaced in-equivalence at the language layer.

### Solution

Implement [Simple Auction example](https://docs.soliditylang.org/en/latest/solidity-by-example.html#simple-open-auction) and add tests to cover all methods

PR #532, opened 2023-10-26
Adding tests for the Simple Auction smart contract.

stefan-stefanooov:
**Description**:
Adding tests for the Simple Auction smart contract.

**Related issue(s)**:
[[Solidity Support] Simple auction example#489](https://github.com/hashgraph/hedera-smart-contracts/issues/489)


test/solidity/simple-auction/simple-auction.js
```diff
@@ -0,0 +1,118 @@
...
+    expect(currentContractBalance, "The contract balance to be the combined of the two transactions").to.equal(combined)
+
+    // Call the withdraw function with the previous highest bidder's address
+    const withdrawTx = await contract.connect(wallet).withdraw()
+    await withdrawTx.wait()
+  
+    await sleep(1000)
+    // Check that the amount of Ether returned to the previous highest bidder is correct
+    const newContractBalance = await ethers.provider.getBalance(contract.address)
+    expect(newContractBalance, "The new balance to be bidAmount").to.equal(bidAmount)
```
konstantinabl, line 103: Maybe you can also test if the balance of the wallet has increased comapred to before the withdraw?

test/solidity/simple-auction/simple-auction.js
```diff
@@ -0,0 +1,118 @@
...
+    expect(initialEvent).to.equal('HighestBidIncreased')
+  })
+
+  it('should confirm bid not high enough scenario works: BidNotHighEnough', async function () {
+    try {
+      const tr = await contract.bid({ value: bidAmountSmall })
+      const receipt = await tr.wait();
+      expect(receipt).to.not.be.null
+    } catch (error) {
+      hasError = true
```
konstantinabl, line 56: I am not sure this is a best practice/good way of writing test - using try/catch and hasError boolean value. I know we do it in some places, but does not seem the best way to do it. Personally, I use chaiAsPromised, you can see and example in my PR for the Blind Auction example - https://github.com/hashgraph/hedera-smart-contracts/blob/02b716d6a4a81d6cc83980c095931c420f311ce2/test/solidity/blind-auction/blindAuction.js#L203

@georgi-l95 @Nana-EC @natanasow @Ivo-Yankov what do you guys think?

Nana-EC (approved):

test/solidity/simple-auction/simple-auction.js
```diff
@@ -0,0 +1,123 @@
...
+    wallet = signers[0]
+    wallet1 = signers[1]
+
+    factory = await ethers.getContractFactory(Constants.Contract.SimpleAuction)
+    contractShortLived = await factory.deploy(CONTRACT_SHORT_DURATION, wallet1.address)
+    await contractShortLived.deployed();
+
+    bidAmount = ethers.utils.parseUnits(TRANSACTION_VALUE, 'gwei')
+    bidAmountSmall = ethers.utils.parseUnits(TRANSACTION_VALUE_SMALL, 'gwei')
+    await sleep(5000)
```
Nana-EC, line 47: Q: why all these sleeps?

Nana-EC (commented):
LG but are the sleeps necessary?

test/solidity/simple-auction/simple-auction.js
stefan-stefanooov: I have two contracts deployed here. One is normal, the second is with a very short duration. After the next block it should be expired so I can test the expiration mechanism in the contract. 
That is why I am waiting for the next block to be mined before I start my tests. I am giving time for the `contractShortLived` to expire.

Nana-EC: Can we do this in a better way that sleeping.
Why not poll for a certain block. Sleeps are arbitrary and don't scale.
Poll for either a transaction or a block of desire then do the logic

Nana-EC (changes requested):
Please explore polling logic instead of sleeps
This might be an item to have that's shared for all tests.
@ebadiere may have explored this also

artifacts/contracts/solidity/simple-auction/SimpleAuction.sol/SimpleAuction.dbg.json
```diff
@@ -0,0 +1,4 @@
+{
```
Nana-EC, line 1: No need to check this file in

Nana-EC (approved):
LG, no need to check in the `dbg.json` file

PR #563, opened 2023-11-03
feat: added smart contract to showcase blind auction example (#490)

### LINKED ISSUE #490: [Solidity Support] Blind auction example
### Problem

Occasionally items have shown up that surfaced in-equivalence at the language layer.

### Solution

Implement [Blind auction example](https://docs.soliditylang.org/en/latest/solidity-by-example.html#id2) and add tests to cover all methods

### Alternatives

_No response_

konstantinabl:
**Description**:

Adds tests for the Blind Auction example from the solidity docs
**Related issue(s)**:

Fixes #490


test/solidity/blind-auction/blindAuction.js
```diff
@@ -0,0 +1,205 @@
...
+    const contract = await deployBlindAuctionContract(9, 5, beneficiary.address);
+
+    const firstBid = ethers.utils.solidityKeccak256(["uint256", "bool", "uint256"] ,[hundredHbars, false, ethers.utils.formatBytes32String('2')]);
+    const secondBid = ethers.utils.solidityKeccak256(["uint256", "bool", "uint256"] ,[twoHundredHbars, true, ethers.utils.formatBytes32String('23')]);
+    const thirdBid = ethers.utils.solidityKeccak256(["uint256", "bool", "uint256"] ,[twoHundredHbars, false, ethers.utils.formatBytes32String('5')]);
+
+    const bid = await contract.connect(wallet1).bid(firstBid, {value: hundred});
+    await bid.wait();
+
+    await sleep(2000);
```
georgi-l95, line 108: Why are we sleeping here and in all other places? Shouldn't wait be sufficient?

test/solidity/blind-auction/blindAuction.js
```diff
@@ -0,0 +1,205 @@
...
+    await contract.deployed()
+
+    return contract;
+}
+
+describe('Solidity Errors', function () {
+  let beneficiary, wallet1;
+  const hundred = ethers.utils.parseEther("100.0");
+  const twoHundred = ethers.utils.parseEther("200.0");
+  const fiftyGwei = ethers.utils.parseUnits('50', 'gwei');
```
natanasow, line 43: These are not meaningful in our case. Let's define them with names like 10 HBARs, 50 HBARs, and so on. Also, the calculation should be based on weibars coef and tinybar-hbar coef.
```typescript
const WEIBARS_COEF = 10_000_000_000;
const TINYBAR_TO_HBAR_COEF = 100_000_000;
const TEN_HBARS = 10 * WEIBAR_COEF * TINYBAR_TO_HBAR_COEF;
```
These coef could be found in one of the constant files, I guess.

natanasow (approved):
LGTM, one nit

test/solidity/blind-auction/blindAuction.js
natanasow: Agreed.

Nana-EC: Let's not sleep as it's usually a sign we're missing something.
Waiting on the transaction hash should ensure the transaction has hit the mirror node and that the state changes it's caused are persisted.
If that's not the case let's investigate the issue

Nana-EC (changes requested):
Logic is good but let's remove all the sleeps and rely on the awaits

test/solidity/blind-auction/blindAuction.js
konstantinabl: @georgi-l95  @Nana-EC  @natanasow 
Okay so I removed the sleeps in this test except for the one here -> https://github.com/hashgraph/hedera-smart-contracts/blob/eaef6fe553e87f2b3c653b6a78a074094018d686/test/solidity/blind-auction/blindAuction.js#L116 
Which i reduced to 1 second - we need it because sometimes bidding takes less than the seconds I have set for biddingEnd so it will fail
The other sleep that seems necessary is the one after the withdraw https://github.com/hashgraph/hedera-smart-contracts/blob/eaef6fe553e87f2b3c653b6a78a074094018d686/test/solidity/blind-auction/blindAuction.js#L131
Behaviour is inconsitent if i remove it sometimes the first expect fails apparently because the balance has not been updated. When I add sleep of 1 seconds it always passes

Nana-EC: I still don't like the sleep. Every transaction submitted to the relay can be awaited and it ensures the transaction went E2E so there should be no need for sleeping.
Are we not awaiting on the right execution

georgi-l95: I think i've seen the same issue with the balances somewhere else, could this be a more complex bug with the relay itself ?

test/solidity/blind-auction/blindAuction.js
konstantinabl: @natanasow done :) good sueggestion

test/solidity/blind-auction/blindAuction.js
Nana-EC: Can we open a ticket to circle back and remove these sleeps

Nana-EC (dismissed):

ebadiere (approved):
LG.

test/solidity/blind-auction/blindAuction.js
konstantinabl: @Nana-EC I removed the ones that are not unecessary, the ones I have left out are needed
It is part of the business logic of the contract. Is there another way to wait for a few block to pass?

PR #567, opened 2023-11-03
update: renamed methods in ControlStructures contract (#566)

### LINKED ISSUE #566: ControlStructures breaks Foundry test in CI
### Problem

Currently, methods in `ControlStructures` contract all start with "test", which is a special keyword set in `foundry-rs/forge` testing framework. While ControlStructures' methods are not meant for forge testing, all these methods start with "test", thus, get picked up by the `forge test` command hence causing the disruptions in CI.

### Solution

- Renamed methods' names

### Alternatives

_No response_

quiet-node:
**Description**:
This PR renamed the methods in ControlStructures contract by replacing the special keyword "test" with "evaluate" so that the `forge test` wouldn't pick up these methods hence wouldn't cause disruptions in foundry-test CI 

**Related issue(s)**:

Fixes #566

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/solidity/control/ControlStructures.sol
+++ b/contracts/solidity/control/ControlStructures.sol
@@ -16,7 +16,7 @@ contract ControlStructures {
         testContract = new TestTryCatchContract();
     }
 
-    function testIfElse(bool condition) external pure returns(bool) {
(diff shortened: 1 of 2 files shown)


Nana-EC (approved):
LG
Nice catch

PR #569, opened 2023-11-03
feat: added CrowdFund solidity OZ Ownable example smart contract (#547)

### LINKED ISSUE #547: [OZ 5.x Contract Support] Add OZ ownable example
### Problem

The Hedera network supports smart contracts as part of EVM equivalence.
Occasionally items have shown up that surfaced inequivalence at the language or opcode layer.

OZ provides many useful templates that smart contract devs are used to utilizing.
However, it may not be clear what coverage the Hedera Smart Contract Service 2.0 has for them

### Solution

- Add solidity contract example that showcase the OZ ownable contract's capabilities

- Add a contract under contracts/OZ/Ownable and tests under respective directories

Ref: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol

### Alternatives

_No response_

quiet-node:
**Description**:
This PR adds CrowdFund contract to showcase the capabilities of the OZ Ownable contract

**Related issue(s)**:

Fixes #547 

**Notes for reviewer**:

**Checklist**


test/OZ/ownable/CrowdFund.js
```diff
@@ -0,0 +1,164 @@
...
+
+    // assertions
+    expect(funderAddress).to.eq(await funder.getAddress())
+    expect(fundedAmount).to.eq(Math.round(FUND_AMOUNT / TINY_BAR_TO_WEI_COEF))
+    expect(contractBalance).to.eq(
+      Math.round(FUND_AMOUNT / TINY_BAR_TO_WEI_COEF)
+    )
+  })
+
+  it('Should allow owner to withdraw an amount which is less than contract balance', async () => {
```
Nana-EC, line 74: Add a test case for when the owner tries to remove more than the contract balance and confirm failure

Nana-EC (changes requested):
Looks good.
1 suggested test case

test/OZ/ownable/CrowdFund.js
quiet-node: Agree! It's a good case scenario to add!

test/OZ/ownable/CrowdFund.js
```diff
@@ -0,0 +1,188 @@
...
+    await crowdFund.connect(funder).deposit({ value: FUND_AMOUNT })
+
+    // prepare transaction to withdraw an amount by owner
+    const WITHDRAWN_AMOUNT = 40000000000 // > FUND_AMOUNT
+    const tx = await crowdFund
+      .connect(owner)
+      .withdraw(WITHDRAWN_AMOUNT / TINY_BAR_TO_WEI_COEF)
+
+    try {
+      await tx.wait()
```
Nana-EC, line 124: Add an `expect` to ensure there was actually an error thrown.

test/OZ/ownable/CrowdFund.js
```diff
@@ -0,0 +1,188 @@
+/*-
+ *
+ * Hedera JSON RPC Relay - Hardhat Example
```
ebadiere, line 3: Please update to state the correct repo, hedera smart contracts.

test/OZ/ownable/CrowdFund.js
quiet-node: Isn't the line 126 expect the error to be thrown? Sorry I'm a little confused by this.

Nana-EC: Line 126 is fine but what if like 124 doesn't result in an error. 
This test will still pass.
We're expecting an error, if one isn't throw we should fail.
Either insert an expect after line 124 or one outside of the try/catch

quiet-node: Ah yeah you're right that makes sense! Thanks for pointing it out!

Nana-EC (approved):

PR #574, opened 2023-11-07
Adds test for reentrancy guard

### LINKED ISSUE #565: Add OZ ReentrancyGuard example

konstantinabl:
**Description**:

This PR adds test for the OZ ReentrancyGuard contract
**Related issue(s)**:

Fixes #565

### DIFF
--- a/test/solidity/reentrancy/reentrancy.js
+++ b/test/solidity/reentrancy/reentrancy.js
@@ -1,0 +1,74 @@
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2023 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+const { expect } = require('chai')
+const { ethers } = require('hardhat')
+const Constants = require('../../constants')
+
+describe('Reentrancy Guard', function () {
+  const tenHBAR = ethers.utils.parseEther("10.0");
+  async function deployContractsAndSendHbars() {
+    const [owner] = await ethers.getSigners();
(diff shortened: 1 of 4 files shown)


georgi-l95 (approved):
LGTM

test/solidity/reentrancy/reentrancy.js
```diff
@@ -0,0 +1,74 @@
...
+    const [owner] = await ethers.getSigners();
+    const factorySender = await ethers.getContractFactory(
+        Constants.Contract.ReentrancyGuardTestSender
+    )
+    contractSender = await factorySender.deploy()
+    const factoryReceiver = await ethers.getContractFactory(
+        Constants.Contract.ReentrancyGuardTestReceiver
+      )
+    contractReceiver = await factoryReceiver.deploy(contractSender.address)
+    const transactionHash = await owner.sendTransaction({
```
natanasow, line 37: nit: we can add a payable constructor to `ReentrancyGuardTestSender`, fund it on contract deployment, and remove that tx completely

natanasow (dismissed):
LGTM, some nits. 

P.S.: I would have started with a different approach based on a predefined execution count (e.g. 5/10/15), and not count on a contract or account balance as a reentrant condition. But that doesn't mean these tests are not enough. The contracts and tests you have written completely satisfy the test plan :clap: :rocket:.

test/solidity/reentrancy/reentrancy.js
```diff
@@ -0,0 +1,74 @@
...
+    expect(counter).to.eq(10)
+    expect(senderBalance).to.eq(0)
+    expect(receiverBalance).to.eq(tenHBAR)
+  })
+
+  it('should verify it cannot reenter with guard', async function () {
+    await deployContractsAndSendHbars();
+
+    await contractReceiver.setNonReentrant(true);
+    const res = await contractReceiver.attackNonReentrant()
```
Nana-EC, line 63: Q: to make the tests the same shouldn't we also be calling `contractReceiver.attack()`.
Why do we use `contractReceiver.attackNonReentrant()` instead?

konstantinabl: In the first case i call attack on a function without the nonReentrant modifier, the other has the nonReentrant modifier and thus it is called attackNonReentrant. I do this because I want to check both cases :) Hope its clear

konstantinabl: In one case the function that is called does not have the nonReentrant modifier, the other one has it. I wanted to test both cases :) Have I answered your question?

natanasow (approved):
LGTM :clap: :rocket:

Nana-EC (approved):

PR #581, opened 2023-11-08
Solidity support panic errors - remove failing test

konstantinabl:
**Description**:

Removes a test, which is not throwing an error as expected.
When we are using pop on an empty array we epect to receive a panic error.
**Related issue(s)**:

Fixes #


konstantinabl:
@Nana-EC It appears that we are not throwing panic error 0x31 as expected. I assume I should open a bug?

test/solidity/errors/panicErrors.js
```diff
@@ -74,11 +74,11 @@ describe('@solidityequiv3 Panic Errors', function () {
     expect(error.errorArgs).to.deep.eq([ethers.BigNumber.from(33)])
   })
 
-  it('should verify panic error 0x31', async function () {
+  //currently seems like it is not working
+  it.skip('should verify panic error 0x31', async function () {
```
ebadiere, line 78: Please add a comment with the issue number tracking this failing test.

konstantinabl:
The issue I opened for the test https://github.com/hashgraph/hedera-smart-contracts/issues/582

Nana-EC (approved):

Issue #571, opened 2023-11-04
Polling required to run some of the smart contracts tests.


ebadiere:
### Description

When a smart contract updates the DLT state the results are not available immediately, as the transaction takes a couple of seconds to get to the mirror node.  Polling in the tests was added to handle this but it's fragile.  A better approach would be to query for the transaction state using the transaction receipt if possible.

### Steps to reproduce

The utils/helpers.js has the polling functions.  A search through the tests will show where they are used and can be replaced with a better solution.

### Hedera network

other

### Version

v0.6.0

### Operating system

macOS

ebadiere:
When running these tests locally, the polling can be reduced to 10 retries and 3000ms between tries.

PR #583, opened 2023-11-08
Removed extra RETRY_DELAY definition.

ebadiere:
Cleanup.  Related issue, #571 

**Related issue(s)**:

Fixes #


local.env
```diff
@@ -14,7 +14,6 @@ PRIVATE_KEYS=
 0x95eac372e0f0df3b43740fa780e62458b2d2cc32d6a440877f1cc2a9ad0c35cc,
 0x6c6e6727b40c8d4b616ab0d26af357af09337299f09c66704146e14236972106"
 
-RETRY_DELAY=3000 # ms
```
Nana-EC, line 17: SO do we still need this large polling logic or can this be reduced by default?

ebadiere: We do, we just don't need it defined twice.

Nana-EC:
@ebadiere please rebase

ebadiere:
Not needed.

PR #584, opened 2023-11-08
feat: Enhanced retries. Renamed functions.  Bumped down retries to see how it behaves in CI.

ebadiere:
Bumped down retries.

**Related issue(s)**: #571 

Fixes #


utils/helpers.js
```diff
@@ -267,6 +269,17 @@ const unPauseAndPoll = async(ERC20Pausable) => {
 
     return false // paused
 }
+
+const pollForResponseCode = async(contract) => {
+  for (let numberOfTries = 0; process.env.MAX_RETRY <= process.env.MAX_RETRY; numberOfTries++) {
+    const event = contract.events.filter(
+      (e) => e.event === Constants.Events.ResponseCode
+    )[0].args[0]
+    return parseInt(event._hex);
```
Nana-EC, line 278: Q: How does this work?
Won't the function return the on the first iteration every time?
There's no condition to decide if to return or not so i'm not sure how the for loop is providing value.
Did you mean to check for event characteristics?

utils/helpers.js
```diff
@@ -283,5 +296,6 @@ module.exports = {
     pollForNewERC721HollowWalletOwner,
     pollForNewSignerBalance,
     pollForNewWalletBalance,
-    unPauseAndPoll
+    unPauseAndPoll,
+    pollForEvent: pollForResponseCode
```
Nana-EC, line 300: nit: pollForEvent begs the question which event.
Should we leave it as pollForResponseCode since that's what it does?

Nana-EC (changes requested):
Question around the loop logic

utils/helpers.js
ebadiere: Push up the latest.  I am not sure what happened, but it was supposed to have a for loop, and it did on my box.  I may have missed staging the latest version.

utils/helpers.js
ebadiere: I renamed it to `pollForEvent` so that it could be more generic and reused.

test/safe-hts-precompile/SafeHTS.js
```diff
@@ -320,10 +320,8 @@ describe('SafeHTS library Test Suite', function () {
     )
 
     const signers0AfterTokenBalance = await pollForNewBalance(erc20Mock, signer1AccountID, signers0BeforeTokenBalance)
-  
-    const signers1AfterTokenBalance = parseInt(
-      await erc20Mock.balanceOf(signer2AccountID)
-    )
+
+    const signers1AfterTokenBalance = await pollForNewBalance(erc20Mock, signer2AccountID, signers1BeforeTokenBalance)
```
Nana-EC, line 324: `signers1AfterTokenBalance` is never used. Wha was it added for?

utils/helpers.js
```diff
@@ -267,6 +269,19 @@ const unPauseAndPoll = async(ERC20Pausable) => {
 
     return false // paused
 }
+
+const pollForEvent = async(contract) => {
```
Nana-EC, line 273:
```suggestion
const pollForLastEvent = async(contract) => {
```

Nana-EC, line 273: It's unclear was `contract` is, please add a type definition.
In https://github.com/hashgraph/hedera-smart-contracts/pull/584/files#diff-ab063aef47051b3442202d7087747109a96c12399fd2dd7c0c92a48f2c4c32a2R271 you pass in a receipt

Nana-EC, line 273: https://github.com/hashgraph/hedera-smart-contracts/pull/584/files#diff-ab063aef47051b3442202d7087747109a96c12399fd2dd7c0c92a48f2c4c32a2R271 passes 2 params.
I prefer that so we should utilize that param here

Nana-EC (commented):
Some improvement suggestions

test/safe-hts-precompile/SafeHTS.js
ebadiere: `signers1AfterTokenBalance` is used on line 344 in an `expect` validation:
`    expect(signers1AfterTokenBalance).to.equal(
      signers1BeforeTokenBalance + transferredAmount
    )
`

utils/helpers.js
ebadiere: Done.

ebadiere: Yes, renamed to the more appropriate `transactionReceipt`.  It's a JS project so I can't declare an explicit object type.

test/safe-hts-precompile/SafeHTS.js
Nana-EC: Read the git change as a new variable. Seems it existed and was used already

utils/helpers.js
```diff
@@ -267,6 +269,19 @@ const unPauseAndPoll = async(ERC20Pausable) => {
 
     return false // paused
 }
+
+const pollForLastEvent = async(transactionReceipt) => {
+  for (let numberOfTries = 0; process.env.MAX_RETRY <= process.env.MAX_RETRY; numberOfTries++) {
+    const event = transactionReceipt.events.filter(
+      (e) => e.event === Constants.Events.ResponseCode
```
Nana-EC, line 276: You still have this hard coded, you should update the method signature to take in a 2nd param and use it

Nana-EC (changes requested):
Almost there.
Update `pollForLastEvent()`

utils/helpers.js
ebadiere: Updated.

Nana-EC (approved):

Nana-EC (changes requested):
Please update PR title and description to match changes

Nana-EC (approved):

Issue #616, opened 2023-12-06
Add Besu local node configuration


stefan-stefanooov:
### Problem

Adding Besu local node configuration.
This is just for convenience of the user.

Currently we have in our hardhat.config.js configurations for 3 networks. (local, testnet, previewnet)
This MR is going to add a 4th one. For a Local Besu instance (if the user of the repo has one)
This local Besu instance is useful for debugging when a problem with our Local Node is identified. 
You can double check our smart contract tests against the pure Besu node. 

### Solution

Add Besu local node configuration

PR #617, opened 2023-12-06
feat: Adding besu local node config

stefan-stefanooov:
**Description**:
Adding Besu local node configuration.
This is just for convenience of the user.

Currently we have in our hardhat.config.js configurations for 3 networks. (local, testnet, previewnet)
This task is going to add a 4th one. For a Local Besu instance (if the user of the repo has one)
This local Besu instance is useful for debugging when a problem with our Local Node is identified. 
You can double check our smart contract tests against the pure Besu node. 

**Related issue(s)**:
Add Besu local node configuration #616

### DIFF
--- a/TEST_SETUP.md
+++ b/TEST_SETUP.md
@@ -38,6 +38,12 @@ At root, create a `.env` file using the `example.env` as the template and fill o
 
 Adjust the `defaultNetwork` field in the [hardhat.config.js](hardhat.config.js) file based on the network your accounts (specified in the .env file) are associated with.
 
+Available Networks:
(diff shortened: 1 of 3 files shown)


stefan-stefanooov:
> Hi @stefan-stefanooov could you please update the description of the PR as well as the issue #616? They are unclear on what the purposes of including the besu local node. Also like @Nana-EC said I think some sort of documentation which showcase the usage and the purpose of the node would be necessary.

Hey Logan (@quiet-node)
I updated the description. Highlight this is just for convenience.

On the second part of your comment. I am not sure what I need to document here. 
The `besu` local node has its own documentation so I some dev wants to bring it up in his local evn they just need to follow it.
Same as the `Hedera Local node`. We don't provide any documentation on how to bring it up in this repo.

Can you or @Nana-EC please elaborate on what exactly needs documenting here?

Issue #478, opened 2023-10-17
Refine Functionality and Logic in System Contract Mocks


mshakeg:
### Problem

* Current implementations only support burning and wiping of a single NFT at index 0.
* Lack of fee consideration logic during transfers.
* Unclear distinctions between functions with identical implementations. This is in relation to the implementation of the `transferToken` and `transferFrom`; and also the `transferNFT` and `transferFromNFT`functions.
* Incorrect handling of max length conditionals regarding UTF-8 encoding and Hedera's restrictions on name and symbol byte lengths.
* The `HederaFungibleToken` and `HederaNonFungibleToken` contracts does not implement the `IHRC` facade interface.

### Solution

* Extend functionality to support multiple NFTs being burnt and wiped.
* Implement logic to handle fixedFees, fractionalFees, and royaltyFees during transfers.
* Investigate and refine the logic differences between functions with identical implementations.
* Correct the max length conditional handling to align with Hedera's restrictions and UTF-8 encoding standards.
* The `HederaFungibleToken` and `HederaNonFungibleToken` contracts should implement the `IHRC` facade interface.

PR #640, opened 2024-01-12
implement IHRC in HTS system contract mocks

### LINKED ISSUE #598: Expand Solidity Version Compatibility Interfaces
### Problem

The current implementation of interfaces such as `IHRC` in the `hedera-smart-contracts` repository is fixed to a specific Solidity version (e.g., `pragma solidity ^0.8.7;`). This version specificity limits compatibility with projects that use other Solidity versions, particularly those on older versions like 0.7.x. As a result, projects using older Solidity versions are forced to reimplement these interfaces to match their environment, leading to unnecessary duplication of effort and potential inconsistencies.

### Solution

To enhance flexibility and broader compatibility, it is proposed to modify the Solidity version declaration in all interfaces within the `hedera-smart-contracts` repository. The version specification should be broadened to include a wider range of Solidity versions. This could be achieved by using something like `pragma solidity >=0.5.0 <0.9.0;`, which would cover a more extensive range of versions while still maintaining a cap to prevent future breaking changes in Solidity from impacting the interfaces.

### Alternatives

_No response_

mshakeg:
**Description**:

This PR extends the `IHRC` interface solidity version support. Additionally, the HTS System Contract mock now implements the `IHRC` interface.

**Related issue(s)**:

Fixes #598 and task 5 in #478

**Notes for reviewer**:

I'm not sure whether `IHRC(htsAddress).isAssociated(evmAddress)` has yet been implemented, but I've added it here regardless.

**Checklist**

### DIFF
--- a/contracts/hts-precompile/IHRC.sol
+++ b/contracts/hts-precompile/IHRC.sol
@@ -1,7 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
-pragma solidity ^0.8.7;
+pragma solidity >=0.4.9 <0.9.0;
+
+interface IERCCommonToken {
+    function balanceOf(address account) external view returns (uint256);
+}
 
 interface IHRC {
     function associate() external returns (uint256 responseCode);
     function dissociate() external returns (uint256 responseCode);
+
+    function isAssociated(address evmAddress) external view returns (bool); // TODO: pending completion of https://hips.hedera.com/hip/hip-719 ??
 }
+
+interface IHRCCommon is IHRC, IERCCommonToken {}
--- a/test/foundry/HederaFungibleToken.t.sol
+++ b/test/foundry/HederaFungibleToken.t.sol
@@ -234,6 +234,48 @@ contract HederaFungibleTokenTest is HederaTokenUtils, HederaFungibleTokenUtils {
         assertEq(success, true, "expected burn to succeed");
     }
 
+    function test_CanAssociateAndDissociateDirectly() public {
+        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](0);
+        address tokenAddress = _createSimpleMockFungibleToken(alice, keys);
+
+        bool success;
+        uint256 amount = 1e8;
+
+        TransferParams memory transferParams = TransferParams({
+            sender: alice,
+            token: tokenAddress,
+            from: alice,
+            to: bob,
+            amountOrSerialNumber: amount
+        });
+
+        TransferParams memory transferFromBobToAlice = TransferParams({
+            sender: bob,
+            token: tokenAddress,
+            from: bob,
+            to: alice,
+            amountOrSerialNumber: amount
+        });
+
+        (success, ) = _doTransferViaHtsPrecompile(transferParams);
+        assertEq(success, false, 'expected transfer to fail since recipient is not associated with token');
+
+        success = _doAssociateDirectly(bob, tokenAddress);
+        assertEq(success, true, 'expected bob to associate with token');
+
+        (success, ) = _doTransferViaHtsPrecompile(transferParams);
+        assertEq(success, true, 'expected transfer to succeed');
+
+        success = _doDissociateDirectly(bob, tokenAddress);
+        assertEq(success, false, 'expected bob to not dissociate with token while postive balance');
+
+        (success, ) = _doTransferViaHtsPrecompile(transferFromBobToAlice);
+        assertEq(success, true, 'expected transfer to succeed');
+
+        success = _doDissociateDirectly(bob, tokenAddress);
+        assertEq(success, true, 'expected bob to dissociate');
+    }
+
     // negative cases
     function test_CannotApproveIfSpenderNotAssociated() public {
         /// @dev already demonstrated in some of the postive test cases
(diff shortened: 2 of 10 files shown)


mshakeg:
Hey @Nana-EC any idea why the PR Title Check still fails?

contracts/hts-precompile/IHRC.sol
```diff
@@ -1,7 +1,15 @@
 // SPDX-License-Identifier: Apache-2.0
-pragma solidity ^0.8.7;
+pragma solidity >=0.4.9 <0.9.0;
```
Nana-EC, line 2: Can't make changes to this file as it's purpose is to serve as a representation of the system contract redirect calls.
Additions would be the result of an accepted HIP modifying the services

contracts/hts-precompile/IHRC.sol
```diff
@@ -1,7 +1,15 @@
...
+
+interface IERCCommonToken {
+    function balanceOf(address account) external view returns (uint256);
+}
 
 interface IHRC {
     function associate() external returns (uint256 responseCode);
     function dissociate() external returns (uint256 responseCode);
+
+    function isAssociated(address evmAddress) external view returns (bool); // TODO: pending completion of https://hips.hedera.com/hip/hip-719 ??
```
Nana-EC, line 12: Not implemented yet unfortunately

Nana-EC (changes requested):
Changes are to a reserved file that should represent released features.
I'd revert the changes the `IHRC.sol`.
Test addition is fine otherwise

Nana-EC:
> Hey @Nana-EC any idea why the PR Title Check still fails?

Not sure.
Can look into it but it's something we can workaround if it's the only issue.

I left other blocking comments

contracts/hts-precompile/IHRC.sol
mshakeg: I see, but why was it ever locked specifically to solidity 0.8.7? interfaces are generally permissive and the other related interfaces e.g. `IHederaTokenService` do specify `>=0.4.9 <0.9.0;`

contracts/hts-precompile/IHRC.sol
mshakeg: no harm in keeping it in as it will eventually?

test/foundry/HederaFungibleToken.t.sol
```diff
@@ -234,6 +234,48 @@ contract HederaFungibleTokenTest is HederaTokenUtils, HederaFungibleTokenUtils {
...
     }
 
+    function test_CanAssociateAndDissociateDirectly() public {
+        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](0);
+        address tokenAddress = _createSimpleMockFungibleToken(alice, keys);
+
+        bool success;
+        uint256 amount = 1e8;
+
+        TransferParams memory transferParams = TransferParams({
```
AlfredoG87, line 244: should we change `transferParams` to `transferFromAliceToBob` to keep consistency and also be more explicit with the test?

test/foundry/HederaFungibleToken.t.sol
```diff
@@ -244,6 +286,54 @@ contract HederaFungibleTokenTest is HederaTokenUtils, HederaFungibleTokenUtils {
...
     }
+
+    function test_CannotRepeatedlyAssociateAndDissociateDirectly() public {
+        IHederaTokenService.TokenKey[] memory keys = new IHederaTokenService.TokenKey[](0);
+        address tokenAddress = _createSimpleMockFungibleToken(alice, keys);
+
+        bool success;
+        uint256 amount = 1e8;
+
+        TransferParams memory transferParams = TransferParams({
```
AlfredoG87, line 297:
```suggestion
        TransferParams memory transferFromAliceToBob = TransferParams({
```

test/foundry/HederaNonFungibleToken.t.sol
```diff
@@ -515,6 +515,68 @@ contract HederaNonFungibleTokenTest is HederaNonFungibleTokenUtils {
...
+
+        mintResponse = _doMintViaHtsPrecompile(mintParams);
+        serialIdU256 = uint64(mintResponse.serialId);
+
+        assertEq(mintResponse.success, true, "expected success since alice is supply key");
+
+        success = _doAssociateDirectly(bob, tokenAddress);
+        assertEq(success, true, 'expected bob to associate with token');
+
+        TransferParams memory transferParams;
```
AlfredoG87, line 547:
```suggestion
        TransferParams memory transferFromAliceToBob;
```

test/foundry/HederaNonFungibleToken.t.sol
```diff
@@ -525,6 +587,75 @@ contract HederaNonFungibleTokenTest is HederaNonFungibleTokenUtils {
...
+            token: tokenAddress,
+            mintAmount: 0
+        });
+
+        mintResponse = _doMintViaHtsPrecompile(mintParams);
+        serialIdU256 = uint64(mintResponse.serialId);
+
+        assertEq(mintResponse.success, true, "expected success since alice is supply key");
+
+        TransferParams memory transferParams;
```
AlfredoG87, line 616:
```suggestion
        TransferParams memory transferFromAliceToBob;
```

AlfredoG87 (commented):
LG. leaved some nits suggestions.

contracts/hts-precompile/IHRC.sol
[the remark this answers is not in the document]
mshakeg: potentially yes

mshakeg: @Nana-EC is your issue with the updated solidity version? or with the added interfaces to this file? The IHRC interface is compatible will all versions in the updated range so that shouldn't be an issue, but I can move the added interfaces into a separate file.

Nana-EC: @mshakeg Ah no issue with the solidity version. 
We can modify that for compatibility.
Issue is with adding more code to it that doesn't reflect what the consensus nodes actually support.

contracts/hts-precompile/IHRC.sol
Nana-EC: The opposite actually, we share this file and others as the implemented feature set, adding it here gives the wrong impression of support for developers that may take the interface and consume it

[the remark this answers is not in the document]
Nana-EC: Yep, should remove it entirely from this file

ebadiere (approved):
LG.

acuarica (approved):
lg

mshakeg:
Hey @Nana-EC can we get this merged? don't want to have to resolve any potential conflicts.

Nana-EC (approved):

PR #649, opened 2024-01-29
feat: add workflow which allows to start tests against desired tag

### LINKED ISSUE #648: Add workflow which allows to start tests against desired tag
Add workflow which allows to start tests against desired tag for network-node, mirror-node and relay.

Blocked by:
- https://github.com/hashgraph/hedera-local-node/issues/525

georgi-l95:
**Description**:
This PR aims to add manual workflow, which can be triggered by specifying tag which the tests should run against.

**Related issue(s)**:

Fixes #648 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/manual-testing.yml
+++ b/.github/workflows/manual-testing.yml
@@ -1,0 +1,29 @@
+name: Manual Testing
+on:
+  workflow_dispatch:
+    inputs:
+      networkNodeTag:
+        description: 'Specify desired Network Node image tag'
+        required: true
+        default: ''
+      mirrorNodeTag:
+        description: 'Specify desired Mirror-Node image tag'
+        required: true
+        default: ''
+      relayTag:
+        description: 'Specify desired Hedera JSON-RPC Relay tag'
+        required: true
+        default: ''
+run-name: Manual Test with network tag ${{ inputs.networkNodeTag }} mirror node tag ${{ inputs.mirrorNodeTag }} and relay tag {{ inputs.relayTag }}
+jobs:
+  ERC20Contract:
+    name: ERC20 Contract Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: ERC20
+
+  ERC721Contract:
+    name: ERC721 Contract Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: ERC721
--- a/.github/workflows/manual-testing.yml
+++ b/.github/workflows/manual-testing.yml
@@ -146,3 +146,48 @@ jobs:
       networkTag: ${{inputs.networkNodeTag}}
       mirrorTag: ${{inputs.mirrorNodeTag}}
       relayTag: ${{inputs.relayTag}}
+
+  SolidityEquivalence1:
+    name: Solidity Equivalence Suite 1
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: 'solidityequiv1'
+      networkTag: ${{inputs.networkNodeTag}}
+      mirrorTag: ${{inputs.mirrorNodeTag}}
+      relayTag: ${{inputs.relayTag}}
+
+  SolidityEquivalence2:
+    name: Solidity Equivalence Suite 2
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: 'solidityequiv2'
+      networkTag: ${{inputs.networkNodeTag}}
+      mirrorTag: ${{inputs.mirrorNodeTag}}
+      relayTag: ${{inputs.relayTag}}
+
+  SolidityEquivalence3:
+    name: Solidity Equivalence Suite 3
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: 'solidityequiv3'
+      networkTag: ${{inputs.networkNodeTag}}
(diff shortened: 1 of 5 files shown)


.github/workflows/test-workflow.yml
```diff
@@ -47,7 +50,7 @@ jobs:
           version: nightly
 
       - name: Start the local node
-        run: npx hedera start -d --network local
+        run: npx hedera start -d --network-tag=${{inputs.networkTag}} --mirror-tag=${{inputs.mirrorTag}} --relay-tag=${{inputs.relayTag}} --verbose=trace
```
quiet-node, line 53: Question: Above I saw the networkTag, mirrorTag, relayTag are marked required:false, does that mean if those tags are not entered, local node will pick up the latest tags?

quiet-node: Nice work!

georgi-l95: My idea with this manual workflow is to be triggered, when for example we have a new network tag. Instead of checking out repos locally, starting local-node and running tests, we can do it here with a simple click.

quiet-node: Nice initiative!

.github/workflows/manual-testing.yml
```diff
@@ -0,0 +1,148 @@
...
+      relayTag: ${{inputs.relayTag}}
+
+  PrngSystemContract:
+    name: PrngSystemContract Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: PrngSystemContract
+      networkTag: ${{inputs.networkNodeTag}}
+      mirrorTag: ${{inputs.mirrorNodeTag}}
+      relayTag: ${{inputs.relayTag}}
```
quiet-node, line 148: Do you think we should make some room for the Solidity, Yul, and OZ contracts as well? https://github.com/hashgraph/hedera-smart-contracts/blob/main/.github/workflows/solidity-tests.yml

.github/workflows/test-workflow.yml
Nana-EC, line 53: Q: is the `--network local` no longer needed? Maybe local is the default?

.github/workflows/manual-testing.yml
Nana-EC: Might be too much for this case but not a bad idea.
@georgi-l95 have you tried the run out?
How long does everything take already and then when you add the above suggestions?

Nana-EC (commented):
Nice addition.
A few questions

.github/workflows/manual-testing.yml
```diff
@@ -0,0 +1,148 @@
...
+      mirrorNodeTag:
+        description: 'Specify desired Mirror-Node image tag'
+        required: true
+        default: ''
+      relayTag:
+        description: 'Specify desired Hedera JSON-RPC Relay tag'
+        required: true
+        default: ''
+run-name: Manual Test with N:${{ inputs.networkNodeTag }}, M:${{ inputs.mirrorNodeTag }} and R:${{ inputs.relayTag }}
+permissions:
```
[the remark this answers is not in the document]
georgi-l95: Unfortunately it fails otherwise with:

```
Invalid workflow file:  .github/workflows/manual-testing.yml#L19 The workflow is not valid. 
.github/workflows/manual-testing.yml (Line: 19, Col: 3): Error calling workflow '/hedera-smart-contracts/.github/workflows/test-workflow.yml@a7e84d8bd1209ac5886ee6f836d53f7d32333bfb'. 
The workflow is requesting 'checks: write, contents: write, pull-requests: write', but is only allowed 'checks: none, contents: read, pull-requests: none'.
```

More info can be found [here](https://www.kenmuse.com/blog/github-actions-workflow-permissions/).

.github/workflows/manual-testing.yml
georgi-l95: Yep, we can add them, they run in parallel, so it shouldn't run longer than now, which is as long as the slowest test (around 5 minutes)

.github/workflows/test-workflow.yml
georgi-l95: So local was always default, not sure why we explicitly used `--network local`. Current behaviour is you can use `--network local`, you can pass tags as we do it now, and if you pass empty strings instead, like `--network-tag=""`, which will be the case when they run on a PR network will be local (as it was until now).

test/multicall/Multicall.js
```diff
@@ -324,16 +324,16 @@ describe('Multicall Test Suite', function () {
       expect(receipt.status).to.eq(1);
     });
 
-    it('should NOT be able to aggregate 150 calls to processLongInputTx', async function () {
-      const n = 150;
+    it('should NOT be able to aggregate 200 calls to processLongInputTx', async function () {
```
georgi-l95, line 327: this change is needed, because of this [PR](https://github.com/hashgraph/hedera-json-rpc-relay/pull/1978) in the relay, which changes chunk size to around 50kb

quiet-node: Oh is that what it is? No wonder it kept failing in the ci

quiet-node (approved):
LGTM. Thanks for the work!

isavov (approved):
That's a great addition!

Nana-EC (approved):

PR #664, opened 2024-02-06
feat: added negative test for Non existing addr and 'send', 'transfer', 'call' sol functions

stefan-stefanooov:
**Description**:
Adding some tests for calling Non existing address. 
Adding some negative tests for 'send', 'transfer', 'call' transactions

**Related issue(s)**:
https://github.com/hashgraph/hedera-services/issues/4672
https://github.com/hashgraph/hedera-json-rpc-relay/issues/805


quiet-node (approved):
LGTM

PR #666, opened 2024-02-08
feat: added atomicHTS contract to showcase examples highlighting the utilization of the HIP-551 (#650)

### LINKED ISSUE #650: HIP 551 HTS Atomic Examples
### Problem

HIP 551 highlights batch transactions.
During the discussions examples were highlighted that could be captured by Smart Contract Scenarios

### Solution

Add examples of common batch transactions using HTS calls

- Unfreeze -> Transfer -> Freeze
- Wipe -> Mint -> Transfer
- Mint -> Unfreeze -> Transfer -> Freeze
- Associate -> Mint -> Transfer
- Associate -> Transfer

Additional considerations noted in https://github.com/hashgraph/hedera-improvement-proposal/pull/551#discussion_r1189004582 include
Associate -> Contract Call

> The contract call is going to send a token to the user. The user must associate the token first, or the contract will revert.

Allowance Approval -> Contract Call

> The contract call is going to take tokens from a user. The user must grant a token allowance first, or the contract will revert.

Allowance Approval -> Allowance Approval -> Associate -> Contract Call

> The contract call is going to take 2 different tokens from a user and send the user an unassociated token that has already been created on the network. The user must grant token allowances first and associate the receivable token, or the contract will revert.

Allowance Approval -> Allowance Approval -> Account Update -> Contract Call

> The contract call is going to take 2 different tokens from a user and send the user an unassociated token that will be created in the contract call. The user must grant token allowances first and have an available auto association slot, or the contract will revert.

Contract Call -> Contract Call

> First contract call is an external contract. Second contract call is a contract validating the results of the first contract. The second contract will revert if the first contract does not produce the intended results. This could be a security enhancement and improve contract call extensibility.

### Alternatives

_No response_

quiet-node:
**Description**:
- added atomicHTS contract to showcase examples highlighting the utilization of the HIP-551
- updated the maxSupply in the TokenCreateContract wrapper contract to increase the capacity of the token

**Related issue(s)**:

Fixes #650 

**Notes for reviewer**:

**Checklist**


contracts/hts-precompile/AtomicHTS.sol
```diff
@@ -0,0 +1,166 @@
...
+    event BatchDelta(int wipeTokenResponseCode, int mintTokenResponseCode, int transferTokenResponseCode);
+    event BatchEcho(int mintTokenResponseCode, int unfreezeTokenResponseCode, int grantKYCResponseCode, int freezeTokenResponseCode);
+    event BatchFoxtrot(int associateResponseCode, int mintTokenResponseCode, int grantKYCResponseCode, int transferTokenResponseCode);
+
+
+    /**
+     * @dev associates, grant token KYC, and send an amount of fungible token to a receiver.
+     * - associateToken() -> grantTokenKYC() -> transferToken()
+     */
+    function batchAlpha(address token, address sender, address receiver, int64 amount) external {
```
Nana-EC, line 29:
```suggestion
    function batchAssociateGrantKycTransfer(address token, address sender, address receiver, int64 amount) external {
```

Nana-EC: Should name the functions in this manner for explicitness

contracts/hts-precompile/AtomicHTS.sol
```diff
@@ -0,0 +1,166 @@
...
+        (int associateResponseCode) = HederaTokenService.associateToken(receiver, token);
+        require(associateResponseCode == HederaResponseCodes.SUCCESS, "Failed to associate token.");
```
Nana-EC, line 31: Allow for already associated response code also

Nana-EC (commented):
LG
Suggestion on association and better naming of functions

contracts/hts-precompile/AtomicHTS.sol
quiet-node: Hmm sorry not quite clear on this. The .associateToken() only returns one response code which is expected to be 22 after associate the token to the account. How is the `TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT = 194` response code supposed to fit into this scenario?

Nana-EC: Won't [associatedToken(receiver, token)](https://github.com/hashgraph/hedera-smart-contracts/blob/main/contracts/hts-precompile/IHederaTokenService.sol#L373) return `TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT` is the contract was already associated?
If so i'm saying instead of failing the logic should handle `SUCCESS` and `TOKEN_ALREADY_ASSOCIATED_TO_ACCOUNT` since the goal is to make sure the contract is associated. Subsequents runs of the same function should fail if done in this way

contracts/hts-precompile/AtomicHTS.sol
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * @dev associates, grant token KYC, and send an amount of fungible token to a receiver.
+     * - associateToken() -> grantTokenKYC() -> transferToken()
+     */
+    function batchAssociateGrantKYCTransfer(address token, address sender, address receiver, int64 amount) external {
+        (int associateResponseCode) = HederaTokenService.associateToken(receiver, token);
+        require(associateResponseCode == HederaResponseCodes.SUCCESS, "Failed to associate token.");
+
+        /// @notice an account needs to be granted the KYC of the HTS token for it to receive the token
+        (int grantKYCResponseCode) = HederaTokenService.grantTokenKyc(token, receiver);
```
Nana-EC, line 34: Similar concept here. If there's a response code for already granted we should allow that and not fail. Same thing further down

Nana-EC (commented):
Looking good.
Just handle the success or state is already as expected response codes mentioned

contracts/hts-precompile/AtomicHTS.sol
quiet-node: Ahh I see what you are saying! Gotcha pushing the fix soon!

contracts/hts-precompile/AtomicHTS.sol
quiet-node: Pushed in the fix. Also I figured that this concept is only applicable for associate methods. Other actions like grantKYC or freeze can be overlapped.

Nana-EC (approved):
LGTM

ebadiere (approved):
LG.

PR #693, opened 2024-03-04
fix: simplified contract artifacts (#692)

### LINKED ISSUE #692: Simplifying Contract Artifacts
### Problem

Presently, the repository commits contract artifacts to retain contract ABIs for reference by other developers. While this is beneficial, non-critical updates (such as changes in Solidity versions) result in the contract bytecode and deployed bytecode being overwritten. This may lead to potential inconsistencies and difficulties in tracking changes over time. To address this issue, we should consider implementing a solution that preserves only the ABIs and ignore other attributes.

### Solution

- include the artifacts folder in the .gitignore to ignore all contract artifacts generated by compilation when committing.
- utilize the `hardhat-abi-exporter` plugin to generate an ABI folder which contains only ABIs of all the contract

### Alternatives

_No response_

quiet-node:
**Description**:
This PR streamlines contract artifacts by adopting ABI-only artifacts, replacing the default artifact files that include various attributes (_contractName, sourceName, abi, bytecode, deployedBytecode, linkReferences, deployedLinkReferences). Previously, all artifact files were committed to the repository, but now only ABI-only artifacts are included. Default artifacts continue to be generated post-compilation alongside ABI-only artifacts, but they are no longer committed to the repository.

Tasks done: 
- tweaked the .gitignore file to drop the artifacts folder except for the necessary contracts required for the System Contract DApp
- installed `hardhat-abi-exporter` plugin to generate ABI-only artifacts
- configured `hardhat-abi-exporter` in the `hardhat.config.js` file to allow custom output path and run on every compilation
- added the folder contains ABI-only artifacts to the commit tree

**Related issue(s)**:

Fixes #692

**Notes for reviewer**:

**Checklist**


Nana-EC (changes requested):
Good change but too large a PR.
Please break this up

ebadiere (approved):
LG.

PR #698, opened 2024-03-05
feat: added hardhat-abi-exporter plugin to generate ABI-only artifacts on compile

### LINKED ISSUE #692: Simplifying Contract Artifacts

quiet-node:
**Description**:
- installed hardhat-abi-exporter plugin
- configured the plugin to automatically generate ABI-only artifacts on compile
- added a new folder `contracts-abi` which contains the new ABI-only artifacts

**Related issue(s)**:

Partially Fixes #692

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ artifacts/contracts/**/**/*.dbg.json
 .openzeppelin/unknown-298.json
 .env
 test-results.*
+contracts-abi/*
+!contracts-abi/contracts
 
 ## --- Foundry Gitignore ---
 
--- a/contracts-abi/contracts/diamond-pattern/Diamond.sol/Diamond.json
+++ b/contracts-abi/contracts/diamond-pattern/Diamond.sol/Diamond.json
@@ -1,0 +1,103 @@
+[
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "_contractOwner",
+        "type": "address"
+      },
+      {
+        "internalType": "address",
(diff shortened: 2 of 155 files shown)


.gitignore
```diff
@@ -9,6 +9,8 @@ artifacts/contracts/**/**/*.dbg.json
 .openzeppelin/unknown-298.json
 .env
 test-results.*
+contracts-abi/*
+!contracts-abi/contracts
```
acuarica, line 13: small nit, does it make sense to exclude OpenZeppelin's ABI directly?

```suggestion
contracts-abi/@openzeppelin
```

acuarica (approved):
lgtm

.gitignore
quiet-node: I think the main purpose is to showcase the ABIs of contracts in our repo. So I think leaving OZ's ABI out won't really be a big deal.

PR #701, opened 2024-03-07
chore: cleaned up Solidity warnings

### LINKED ISSUE #435: Warnings when compiling contracts can be cleaned up.
### Description

Compiling contracts produces warnings which does not look professional.  They can be cleaned up.

### Steps to reproduce

Most of the warnings are of unused variables.  These variables can most likely be removed.

### Additional context

_No response_

### Hedera network

other

### Version

v0.5.0

### Operating system

macOS

quiet-node:
**Description**:
- cleaned up all warnings thrown by Solidity after compilation. 
- fixed unit tests where needed
- there're only 4 warnings left but only because they serve the purpose of the coverages

```bash
Warning: This declaration shadows an existing declaration.
  --> contracts/solidity/scoping/Scoping.sol:21:13:
   |
21 |             uint x;
   |             ^^^^^^
Note: The shadowed declaration is here:
  --> contracts/solidity/scoping/Scoping.sol:18:9:
   |
18 |         uint x = 1;
   |         ^^^^^^


Warning: "selfdestruct" has been deprecated. The underlying opcode will eventually undergo breaking changes, and its use is not recommended.
  --> contracts/discrepancies/nonce/InternalCallee.sol:27:9:
   |
27 |         selfdestruct(_addr);
   |         ^^^^^^^^^^^^


Warning: "selfdestruct" has been deprecated. The underlying opcode will eventually undergo breaking changes, and its use is not recommended.
  --> contracts/discrepancies/nonce/InternalCaller.sol:56:9:
   |
56 |         selfdestruct(_addr);
   |         ^^^^^^^^^^^^


Warning: "selfdestruct" has been deprecated. The underlying opcode will eventually undergo breaking changes, and its use is not recommended.
  --> contracts/solidity/signature-example/ReceiverPays.sol:27:9:
   |
27 |         selfdestruct(payable(msg.sender));
   |         ^^^^^^^^^^^^

```

**Related issue(s)**:

Fixes #435 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/discrepancies/nonce/InternalCallee.sol
+++ b/contracts/discrepancies/nonce/InternalCallee.sol
@@ -15,14 +15,15 @@ contract InternalCallee {
     }
 
     function revertWithRevertReason() public returns (bool) {
+        if (calledTimes < 0) {++calledTimes;}
         revert("RevertReason");
     }
 
-    function revertWithoutRevertReason() public returns (bool) {
+    function revertWithoutRevertReason() public pure returns (bool) {
         revert();
     }
 
-    function selfdestruct(address payable _addr) external {
+    function selfDestruct(address payable _addr) external {
         selfdestruct(_addr);
     }
 }
--- a/contracts/discrepancies/nonce/InternalCallee.sol
+++ b/contracts/discrepancies/nonce/InternalCallee.sol
@@ -11,11 +11,13 @@ contract InternalCallee {
     }
 
     function externalFunction() external returns (uint) {
+        // mutate state to maintain non-view function status
         return ++calledTimes;
     }
 
     function revertWithRevertReason() public returns (bool) {
-        if (calledTimes < 0) {++calledTimes;}
+        // mutate state to maintain non-view function status
+        ++calledTimes;
         revert("RevertReason");
     }
 
--- a/contracts/discrepancies/nonce/InternalCaller.sol
+++ b/contracts/discrepancies/nonce/InternalCaller.sol
@@ -5,14 +5,16 @@ contract InternalCaller {
     constructor() payable {}
 
     function callNonExisting(address _addr) external {
-        _addr.call(abi.encodeWithSignature("nonExisting()"));
+        (bool success,) = _addr.call(abi.encodeWithSignature("nonExisting()"));
+        require(success);
     }
 
-    function staticCallNonExisting(address _addr) external {
-        _addr.staticcall(abi.encodeWithSignature("nonExisting()"));
+    function staticCallNonExisting(address _addr) view external {
+        (bool success,) = _addr.staticcall(abi.encodeWithSignature("nonExisting()"));
+        require(success);
     }
 
-    function staticCallExternalFunction(address _addr) external returns (uint) {
+    function staticCallExternalFunction(address _addr) view external returns (uint) {
         (bool success, bytes memory result) = _addr.staticcall(abi.encodeWithSignature("externalFunction()"));
         return success && result.length > 0 ? abi.decode(result, (uint)) : 0;
     }
@@ -28,15 +30,17 @@ contract InternalCaller {
     }
 
     function callRevertWithRevertReason(address _addr) external {
-        _addr.call(abi.encodeWithSignature("revertWithRevertReason()"));
+        (bool success,) = _addr.call(abi.encodeWithSignature("revertWithRevertReason()"));
+        require(success);
     }
 
     function callRevertWithoutRevertReason(address _addr) external {
-        _addr.call(abi.encodeWithSignature("revertWithoutRevertReason()"));
+        (bool success,) = _addr.call(abi.encodeWithSignature("revertWithoutRevertReason()"));
+        require(success);
     }
 
     function sendTo(address payable _addr) external {
-        _addr.send(1);
+        _addr.transfer(1);
     }
 
     function transferTo(address payable _addr) external {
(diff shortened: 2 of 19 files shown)


contracts/discrepancies/nonce/InternalCaller.sol
```diff
@@ -28,26 +30,29 @@ contract InternalCaller {
...
     }
 
     function callRevertWithoutRevertReason(address _addr) external {
-        _addr.call(abi.encodeWithSignature("revertWithoutRevertReason()"));
+        (bool success,) = _addr.call(abi.encodeWithSignature("revertWithoutRevertReason()"));
+        require(success);
     }
 
     function sendTo(address payable _addr) external {
-        _addr.send(1);
```
Nana-EC, line 39: Q: is this a replacement due to warning?
Is there a way to ignore the warning and maintain the coverage?
Maybe even add support for the `transfer` also

contracts/solidity/block/BlockInfo.sol
```diff
@@ -32,7 +32,8 @@ contract BlockInfo {
     }
 
     // should behave like prevrandao
+    /// @notice since VM version Paris, "difficulty" was replaced by "prevrandao", which now returns a random number based on the beacon chain
     function getBlockDifficulty() external view returns (uint256) {
-        return block.difficulty;
+        return block.prevrandao;
```
Nana-EC, line 37: Q: is this a replacement due to warning?
Is there a way to ignore the warning and maintain the coverage?
Also add support for `prevrandao`

contracts/solidity/precompiles/Precompiles.sol
```diff
@@ -5,6 +5,7 @@ contract Precompiles {
 
     event DebugBytes(bytes data);
     event DebugUint256(uint256 value);
+    uint256 constant dummy = 0;
```
Nana-EC, line 8: Q: where's `dummy` used?

contracts/solidity/scoping/Scoping.sol
```diff
@@ -18,7 +18,6 @@ contract Scoping {
         uint x = 1;
         {
             x = 2; // this will assign to the outer variable
-            uint x;
```
Nana-EC, line 21: I believe this is required to showcase that what looks like a declaration of a default `uint` doesn't affect the value of `x`

Nana-EC (changes requested):
Good changes based on warnings but let's make sure we're not losing overage and breaking the aim of the tests

contracts/discrepancies/nonce/InternalCaller.sol
quiet-node: Yes it's due to the warning but I read it too quickly and replaced it with `transfer` but it indeed requires its returned var to be used. Pushed in the fix!

contracts/solidity/block/BlockInfo.sol
quiet-node: Yes, it's because of the warning. Solidity doesn't inherently support ignoring warnings. I did some research on a third-party Hardhat plugin `hardhat-ignore-warning`, but it only works for completely turning off warnings for the entire project, which isn't a good solution. The plugin also offers a function to ignore warnings for specific lines using inline comments, but they don't seem to work. I'm reaching out to the plugin maintainer to ask about this issue. For now, we can revert back to using `block.difficulty` to maintain coverage. And we already have coverage for `prevrandao`." Pushed in the fix.

contracts/solidity/precompiles/Precompiles.sol
quiet-node: Right I was adding it attempting to remove a warning within the smart contract but didn't need it somehow and ended up forgetting about it. Great catch! Pushed in the fix!

contracts/solidity/scoping/Scoping.sol
quiet-node: I removed it due to a warning, then I reached out to Stefan and he sent me the link to the function on Soliditylang website and you're right this line is intended to be added here. Pushed in the fix!

contracts/discrepancies/nonce/InternalCallee.sol
```diff
@@ -15,14 +15,15 @@ contract InternalCallee {
     }
 
     function revertWithRevertReason() public returns (bool) {
+        if (calledTimes < 0) {++calledTimes;}
```
Nana-EC, line 18: Please add a comment to `calledTimes` so it's clear how it's intended to be used.
Right now it's not clear why `externalFunction()` and `revertWithRevertReason()` increment it but other functions don't

Nana-EC (commented):
Looking good.
Just a bit more info needed

contracts/discrepancies/nonce/InternalCallee.sol
quiet-node: The original warning was "Function state mutability can be restricted to pure", then if I made this function a pure function, the nonce of the caller won't be incremented when this method is called which is a case scenario in the UTs. That's why I added the line `if (calledTimes < 0) {++calledTimes}` which is an impossible condition but to make the compiler treat this as a non-view function. Hmm now that I look at this again this line maybe not needed to be here but we can fix the UT, or maybe the case scenario in UTs is the intended case. I'll reach out to Nikolay about this

quiet-node: Reached out to Nikolay and he confirmed that we want to maintain the non-view status of the function so we can just simply add `++calledTimes;`. I pushed in the fix along with comments showing the intended purpose of the state mutation.

Nana-EC (approved):

PR #703, opened 2024-03-13
test: extend the coverage of account nonce discrepancies tests

### LINKED ISSUE #1016: Finish the implementation for PR #703 related to account nonce discrepancies
### Problem

There is this old [PR](https://github.com/hashgraph/hedera-smart-contracts/pull/703) that adds additional tests for account nonce discrepancies and that has to be checked, rebased and updated. 

### Solution

Have a look on the pr and update it so it can be successfully merged

### Alternatives

_No response_

nickeynikolovv:
**Description**:

This PR extends the coverage of account nonce discrepancies tests according to test plan - https://www.notion.so/limechain/Address-account-nonce-discrepancies-Test-Plan-4a186e8cfabc49888ba501750f68d8de?pvs=4

**Related issue(s)**:

Fixes #1016


artifacts/contracts/discrepancies/nonce/ChainedContracts.sol/ChainedContracts.json
```diff
@@ -0,0 +1,55 @@
+{
```
quiet-node, line 1: these artifacts should not be committed to the project after this commit https://github.com/hashgraph/hedera-smart-contracts/commit/dfaedf93e5dc857ce2451921c967230099e43ac5. Please kindly rebase main!

quiet-node, line 1: These artifacts should not be added anymore after this commit https://github.com/hashgraph/hedera-smart-contracts/commit/dfaedf93e5dc857ce2451921c967230099e43ac5 (we only want contract-abi/). Please kindly rebase main and uncheck these artifacts from the git tree

contracts/discrepancies/nonce/InternalCallee.sol
```diff
@@ -1,28 +1,50 @@
 // SPDX-License-Identifier: Apache-2.0
 pragma solidity ^0.8.0;
 
-contract Sample {}
+contract Sample {
+  function selfdestruct() external {
```
quiet-node, line 5: this will throw
```bash
Warning: This declaration shadows a builtin symbol.
```

Change it like below to avoid the warning
```suggestion
  function selfDestruct() external {
```

contracts/discrepancies/nonce/InternalCallee.sol
```diff
@@ -1,28 +1,50 @@
...
+
+  function revertWithRevertReason() public returns (bool) {
+    revert('RevertReason');
+  }
+
+  function revertWithoutRevertReason() public returns (bool) {
+    revert();
+  }
+
+  function selfdestruct(address payable _addr) external {
```
quiet-node, line 29: this will throw
```bash
Warning: This declaration shadows a builtin symbol.
```

Change it like below to avoid the warning
```suggestion
  function selfDestruct() external {
```

contracts/discrepancies/nonce/InternalCallee.sol
```diff
@@ -1,28 +1,50 @@
...
+
+  function factorySample() external returns (address) {
+    return address(new Sample());
+  }
+
+  function externalFunction() external returns (uint) {
+    return ++calledTimes;
+  }
+
+  function revertWithRevertReason() public returns (bool) {
```
quiet-node, line 21:
```bash
Warning: Function state mutability can be restricted to pure
```

contracts/discrepancies/nonce/InternalCallee.sol
```diff
@@ -1,28 +1,50 @@
...
+    revert('RevertReason');
+  }
+
+  function revertWithoutRevertReason() public returns (bool) {
```
quiet-node, line 25:
```bash
Warning: Function state mutability can be restricted to pure
```

quiet-node (changes requested):
Nice work but some comments

artifacts/contracts/discrepancies/nonce/ChainedContracts.sol/ChainedContracts.json
nickeynikolovv: Done. Removed all json files that were uploaded

quiet-node: So we don't want the `default` artifacts JSON files in the `artifacts/` folder to be checked in, but we still want to keep the `ABI-only artifact files` in the `contract-abi/` folder. Let me know if you have any question as I see this could be a little confusing.

contracts/discrepancies/nonce/InternalCallee.sol
nickeynikolovv: renamed to selfDestructSample because we already have selfDestruct function in the same file

quiet-node (dismissed):
LG! Great work thanks!

test/discrepancies/Nonce.js
```diff
@@ -299,12 +342,586 @@ describe('@discrepancies - Nonce Test Suite', async () => {
...
+    expectIncrementedNonce(snBefore, mnBefore, snAfter, mnAfter);
+  })
+
+  //NONCE-017
+  it('should update nonce after successful ERC721 token call', async function () {
+    erc721Contract = await Utils.deployERC721Contract();
+    tokenAddress = await Utils.createNonFungibleToken(
+      tokenCreateContract,
+      signers[0].address
+    );
```
victor-yanev, line 390: Let's avoid overriding variables which are reused in multiple tests - this creates dependencies between the tests and depending on their execution order the variable might hold a different value. This reduces the readability of code and makes it hard to trace what is happening without knowing the order of the tests or without going into debug mode.

If we are going to reuse a variable let's assign it only in one place (e.g., in a `before()` method or, where possible, directly turn it into a `const`.)

Please check for other such occurrences and try to fix them 🙂. Thank you!

test/discrepancies/Nonce.js
```diff
@@ -299,12 +342,586 @@ describe('@discrepancies - Nonce Test Suite', async () => {
...
+    expectIncrementedNonce(0, 0, servicesIntermediaryContractContractNonce, mirrorNodeIntermediaryContractNonce);
+    expectIncrementedNonce(0, 0, servicesOuterContractNonce, mirrorNodeOuterContractNonce);
+
+    //verify nonce of the inner most contract is 0
+    expectNonIncrementedNonce(0, 0, servicesInnerContractNonce, mirrorNodeInnerContractNonce);
+  });
+
+  //NONCE-023
+  //need to update the get contract nonce - waiting for SDK
+  xit('should update contract nonce for each deployed contract', async function () {
```
victor-yanev, line 597: Just a lil reminder to enable this later (and other tests waiting for the SDK changes)

contracts/discrepancies/nonce/ChainedContracts.sol
```diff
@@ -0,0 +1,56 @@
...
+}
+
+contract ChainedContracts {
+    address public owner;
+    Deploys1Contract  public childContract;
+
+    constructor() {
+        owner = msg.sender;
+        // Deploy Chain within ChildContract constructor
+        //  chainContract = new ChainContract();
```
Nana-EC, line 33: nit: remove and other commented out code

test/hts-precompile/utils.js
```diff
@@ -525,6 +530,12 @@ class Utils {
     return accountInfo.accountId.toString();
   }
 
+  static async getContractInfo(evmAddress, client){
```
Nana-EC, line 533: Can you add a comment to note that the only reason we're using the SDK is to call the consensus node to confirm that there are no discrepancies between CN and MN.

test/discrepancies/Nonce.js
```diff
@@ -299,12 +342,586 @@ describe('@discrepancies - Nonce Test Suite', async () => {
...
+    //verify nonce is 0
+    const snAfterNewCreate = await getServicesNonce(wallet.address);
+    const mnAfterNewCreate = await getMirrorNodeNonce(wallet.address);
+
+    expectNonIncrementedNonce(snAfterNewCreate, mnAfterNewCreate, 0, 0);
+  });
+
+  //NONCE-022
+  //should fix the get contract nonce function - waiting for SDKs
+  xit('should update all nonces when chained deploys of contracts', async function () {
```
Nana-EC, line 553: Q: why is this disabled?

Nana-EC (changes requested):
Looking good, a few comments to address.

test/discrepancies/Nonce.js
nickeynikolovv: There are few tests where we are not able to get nonce of a smart contract from the sdk - this issue: https://github.com/hashgraph/hedera-sdk-js/issues/2192 . Basically we cannot assert that the nonce of a smart contract is correct

test/discrepancies/Nonce.js
nickeynikolovv: Unneeded override is removed

test/hts-precompile/utils.js
nickeynikolovv: done

contracts/discrepancies/nonce/ChainedContracts.sol
nickeynikolovv: done

Nana-EC:
@nickeynikolovv please rebase

test/discrepancies/Nonce.js
webmaster1225: right

quiet-node:
Hello @nickeynikolovv is this PR still relavant? If so please kindly update the PR. Thanks!

test/discrepancies/Nonce.js
```diff
@@ -360,14 +331,357 @@ describe('@discrepancies - Nonce Test Suite', async () => {
...
+        internalCalleeContract.deployViaCreate2(1, {gasLimit: 500000})
+    );
+
+    const snAfter = await getServicesNonce(signers[0].address);
+    const mnAfter = await getMirrorNodeNonce(signers[0].address);
+
+    expectIncrementedNonce(snBefore, mnBefore, snAfter, mnAfter);
+  });
+
+  it('should not update nonces when deploying on an address with an already existing account', async function () {
```
Nana-EC, line 442: nit: test name is a little misleading. I assumed there was a contract deployment but rather it's a hollow account -> full account -> deleted account -> hollow account flow.
The nonces are updated when you execute a transfer. In the other 2 checks you confirm the hollow account nonce is zero.
Just update the test description so it's clear

Nana-EC (approved):
Nice additional scenarios.
Minor comment on test description but the additional coverage is great.
Thanks for following through - signing off to unblock

test/discrepancies/Nonce.js
natanasow: Done.

PR #728, opened 2024-05-07
Adds Maian tool analysis report

nabialek-arianelabs:
Adds Maian tool analysis report and docker files.
Is dependent upon: https://github.com/hashgraph/hedera-smart-contracts/pull/726/files

### DIFF
--- a/tools/maian-analysis/signing_transactions.md
+++ b/tools/maian-analysis/signing_transactions.md
@@ -1,0 +1,26 @@
+### Overview
+
+The point of this document is to analyze how different web3 providers handle the eth_sendTransaction method.
+
+Method [eth_sendTransaction](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_sendtransaction) creates a new message call transaction or a contract creation, if the data field contains code, and signs it using the account specified in "from" field.
+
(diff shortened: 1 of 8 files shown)


tools/maian-analysis/signing_transactions.md
```diff
@@ -0,0 +1,26 @@
+### Overview
```
ebadiere, line 1: What's the value here?  The analysis states that "Hedera does not support `eth_sendTransaction`".

ebadiere:
@nabialek-arianelabs I noticed the dependency on:  [PR726](https://github.com/hashgraph/hedera-smart-contracts/pull/726/files).  Once @acuarica and @arianejasuwienas agree to where these kind of reports should live, then I think this PR should follow that standard.

tools/maian-analysis/signing_transactions.md
arianejasuwienas: eth_sendRawTransaction can be used instead. But it would require us to rewrite part of the MAIAN tool code. We won't be able to use Hedera JSON RPC in MAIAN otherwise.

PR replace with: https://github.com/hashgraph/hedera-smart-contracts/pull/738

PR #731, opened 2024-05-07
feat: added new StateRegistry contract to expose more state-related-scenarios

quiet-node:
**Description**:
This PR adds StateRegistry contract mainly for the purpose of migration testing. Bellow is the list of tasks that have been done:

- added a new StateRegistry contract to expose more state related scnearios including:
   - bool state type
   - uint & int state types
   - address state types
   - contract state types
   - bytes32 state types
   - string state types
   - enum state types
   - integer array state types
   - states that are defined and then deleted

- added test suite with different tags for differenct stages 
   - @pre-migration: test suite that runs with mono code
   - @post-migration-view-function: test suite that runs after migration to mod code and only contains coverage for view functions
   - @post-migration-non-view-function: test suite that runs after migration to mod code and contains coverage for both view and non-view functions

- added a writing to disk feature to capture states being registered during mono code. This on-disk-state-object will later be compared with states on mod code

- added a script that freeze the network node using the sdk
- bumped @hashgraph/sdk package up to 2.44.0 to fix the `FreezeType is not a constructor` error when execute freeze-network-node script. The Freeze transaction has a property FreezeType, which should be set with the 'setFreezeType' method. This method accepts the FreezeType class. The problem is that in @hashgraph/sdk@0.41.0, the FreezeType class is not exported from the SDK. This causes an inability to see the freeze type of the transaction. 

credit: @natanasow

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


scripts/freeze-network-node.js
```diff
@@ -0,0 +1,59 @@
...
+  FreezeTransaction,
+  Client,
+  Timestamp,
+  FreezeType,
+} = require('@hashgraph/sdk');
+const { OPERATOR_ID_A, OPERATOR_KEY_A } = require('../utils/constants');
+
+async function main() {
+  try {
+    const genesisClient = Client.forNetwork({
```
Nana-EC, line 34: nit: add a comment to say assumes a single node network.
This won't work in multinode case. 
No change needed but worth noting

test/state-registry/stateRegistry.js
```diff
@@ -0,0 +1,849 @@
...
+    for (let i = 0; i < ITERATIONS; i++) {
+      describe('should test value types', async function () {
+        it('should test bool', async function () {
+          const randomBool = getRandomInt(0, 1) > 0;
+
+          const tx = await contract.setVarBool(randomBool);
+          await tx.wait();
+
+          const resp = await contract.getVarBool();
+          statesObject['VarBool'] = resp;
```
Nana-EC, line 68: For sets like this are we duplicating the action every time the i loop runs?
If so do tests like "should test bool" and  "should test address" be inside the loop?

contracts/state-registry/StateRegistry.sol
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    function setBalance(address addr, uint256 value) external {
+        balance[addr] = value;
+    }
+
+    function balanceOf(address account) public view returns (uint256) {
+        return balance[account];
+    }
+}
```
Nana-EC, line 280: nit: one option we considered was a get function that returns all state variable values.
This would allow you to make one get call per contract before and after migration making it simpler for the state validation logic.

Nana-EC (approved):
For 1st pass this works.
Let's get a few iterations behind it and we can improve later

contracts/state-registry/StateRegistry.sol
natanasow: another nit: we could add a function that hashes entire contract storage and we just compare the hashes before/after migration, the downside of this approach is that we can't get the differences in an easy manner
```solidity
    function getStateHash() external view returns (bytes32) {
        return
            keccak256(
                abi.encode(
                    abi.encode(
                        varUint8, varUint16, varUint32, varUint64, varUint128, varUint256, 
                        varInt8, varInt16, varInt32, varInt64, varInt128, varInt256
                    ),
                    abi.encode(
                        varBool, varAddress, varContractType, varBytes32, varString, 
                        varIntArrDataAllocBefore, varIntArrDataAllocAfter, varIntArrDataAllocDeleted
                    ),
                    abi.encode(
                        varStringConcat, varStringConcatDeleted,
                        varContractStruct, varContractStructDeleted
                    )
                )
            );
    }
 ```

File: contracts/state-registry/StateRegistry.sol [hunk 4]
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    function deleteVarContractStructDeleted() external {
+        delete varContractStructDeleted;
+    }
+
+    function setBalance(address addr, uint256 value) external {
+        balance[addr] = value;
+    }
+
```
natanasow, line 276: we should add a deleting method
```solidity
    function deleteBalance(address addr) external {
        delete balance[addr];
    }
```

test/hts-precompile/token-create/tokenCreateContract.js
```diff
@@ -289,7 +289,7 @@ describe('TokenCreateContract Test Suite', function () {
     ).to.equal(22);
   });
 
-  describe('Hapi vs Ethereum token create test', function () {
+  describe.skip('Hapi vs Ethereum token create test', function () {
```
Nana-EC, line 292: Q: why are skipping?

quiet-node: Hm strange I really thought I've mentioned this somewhere either on Slack or the description but obviously I didn't. Anyway, this test fail when upgrading the sdk from 2.41 to 2.44. So I skip this test to unblock this PR, and then create a new ticket fix it but then this failure was caught in one of the bump PR https://github.com/hashgraph/hedera-smart-contracts/actions/runs/9113486261/job/25055115772?pr=750#step:10:76. So I'll fix this test in the bump PR and then rebase this PR to remove the skipping.

quiet-node: fixed the test in https://github.com/hashgraph/hedera-smart-contracts/pull/750

Nana-EC:
@quiet-node please rebase

Nana-EC (approved):

Nana-EC (approved):

PR #757, opened 2024-05-18
feat: add upgrade testing workflow

natanasow:
**Description**:

Add a **Pre/Post upgrade Testing** CI workflow that uses the user's selected initial network/mirror/relay versions, target network/mirror/relay versions, pre/post upgrade suite tag, and execute the following flow:
-  runs a node with the initial version
- executes pre-upgrade tests
- freezes the network via FreezeTransaction
- runs a node with a target version (simulates node upgrade between different versions)
- executes post-upgrade tests

Examples:
- running `@pre-migration` and `@post-migration` tests against 0.48.1 and 0.49.7
usage1
- running `@OZERC20` tests against 0.48.1 and 0.49.7 (ensure that the node's upgrade has been successfully executed and the state has not been corrupted)
usage2

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**


isavov (approved):

isavov (dismissed):

jeromy-cannon (approved):

Nana-EC (commented):
LG
Please add a README somewhere.

.github/workflows/upgrade-testing.yml
```diff
@@ -0,0 +1,100 @@
+name: Pre/Post upgrade Testing
```
quiet-node, line 1: This name feels a bit confusing. I suggest renaming it to "Migration Testing" or something similar, as it primarily handles tasks related to migration tests. However, we already have a "Migration Testing" in the state testing PR. Maybe we could scratch the Migration Testing workflow in the PR and keep this one, since this one is more comprehensive with all the customized inputs. What do you think?

quiet-node (commented):
Wonderful work @natanasow 🚀 🔥. Just a comment below ⬇️

quiet-node:
> LG Please add a README somewhere.

Agree to Nana about adding a doc somewhere to describe the purpose of this CI workflow so others who don't work on this matter have a better idea on what's the point of this process. But maybe a README is a little overkilled, I was thinking about a `description` block comment somehwere within the .yml file itself.

.github/workflows/upgrade-testing.yml
natanasow: Yep, our end goal should be to completely remove the `migration-testing.yml` workflow and to use only this one from now on because it's much more customizable. I changed the naming because I wanted to de-couple the tests from the entire migration flow. If you insist, I'm good with reverting to the old namings (Migration Testing).

Nana-EC:
> > > LG Please add a README somewhere.
> 
> > 
> 
> > Agree to Nana about adding a doc somewhere to describe the purpose of this CI workflow so others who don't work on this matter have a better idea on what's the point of this process. But maybe a README is a little overkilled, I was thinking about a `description` block comment somehwere within the .yml file itself.
> 
> 
> 
> @Nana-EC Are you okay with adding a description block instead of a whole README?

@natanasow please do README.
I would like an in depth explanation of the E2E upgrade process so developers can see how local node is being used to simulate upgrade.

Also a section to show people how they can use different tests etc.

You can do this in a follow up PR

.github/workflows/upgrade-testing.yml
quiet-node: Nice one yeah I would vote for the old name, Migration Testing.

quiet-node (approved):
LGTM

natanasow:
Follow-up PR must created with the following issues https://github.com/hashgraph/hedera-smart-contracts/issues/770 :warning:

PR #771, opened 2024-05-23
chore: add Solidity files for the Hedera Account Service

lukelee-sl:
**Description**:
Add Solidity files for supporting the Hedera Account Service system contract


david-bakin-sl (approved):
LGTM

david-bakin-sl (approved):
LGTM

PR #775, opened 2024-05-24
BLS Signature Verification Example

AlfredoG87:
**Description**:
added example contract for verifying a BLS Signature using Keccak hashing algorithm

Added test suite, and added test suite to CI. 

Also added some missing tests from the publish results.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts-abi/contracts/bls-signature/Migrations.sol/Migrations.json
+++ b/contracts-abi/contracts/bls-signature/Migrations.sol/Migrations.json
@@ -1,0 +1,59 @@
+[
+  {
+    "inputs": [],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "inputs": [],
+    "name": "last_completed_migration",
+    "outputs": [
+      {
+        "internalType": "uint256",
+        "name": "",
+        "type": "uint256"
+      }
+    ],
+    "stateMutability": "view",
+    "type": "function"
+  },
+  {
+    "inputs": [],
+    "name": "owner",
+    "outputs": [
+      {
+        "internalType": "address",
+        "name": "",
+        "type": "address"
+      }
+    ],
+    "stateMutability": "view",
+    "type": "function"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "uint256",
+        "name": "completed",
+        "type": "uint256"
+      }
+    ],
+    "name": "setCompleted",
+    "outputs": [],
+    "stateMutability": "nonpayable",
+    "type": "function"
+  },
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "new_address",
+        "type": "address"
+      }
+    ],
+    "name": "upgrade",
+    "outputs": [],
+    "stateMutability": "nonpayable",
+    "type": "function"
+  }
+]
--- a/contracts/bls-signature/BLS.sol
+++ b/contracts/bls-signature/BLS.sol
@@ -1,0 +1,29 @@
+pragma solidity ^0.8.0;
+
+import "./Pairing.sol";
+
+/**
+ * @title BLS
+ * @dev A library for verifying BLS signatures. Based on the work of https://gist.github.com/BjornvdLaan/ca6dd4e3993e1ef392f363ec27fe74c4, which is licensed under Apache 2.0 (https://www.apache.org/licenses/LICENSE-2.0).
+ */
+library BLS {
+    /*
+     * Internal functions
+     */
+
+    /**
+     * @dev Checks if a BLS signature is valid.
+     * @param _verificationKey Public verification key associated with the secret key that signed the message.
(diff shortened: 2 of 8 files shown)


test/bls-signature/bls-signature-verification.js
```diff
@@ -0,0 +1,86 @@
...
+
+  before(async function () {
+    signers = await ethers.getSigners();
+    const BLSTestFactory = await ethers.getContractFactory(
+      Constants.Contract.BLSTest
+    );
+
+    BLS = await BLSTestFactory.deploy();
+
+    console.log(
```
quiet-node, line 40: Do we need this log? I know it's for debugging purpose but probably we can have another deployment unit test and make sure that the BLS contract is deployed by expect its address to exist and not be zero address?

test/bls-signature/bls-signature-verification.js
```diff
@@ -0,0 +1,86 @@
...
+    );
+  });
+
+  it('should verify a valid signature', async () => {
+    let signatureX =
+      '11181692345848957662074290878138344227085597134981019040735323471731897153462';
+    let signatureY =
+      '6479746447046570360435714249272776082787932146211764251347798668447381926167';
+    let result = BLS.verify(message, signatureX, signatureY);
+    console.log('Transaction hash (valid signature):', result.tx);
```
quiet-node, line 51: Instead of logging, we should make assertion.

test/bls-signature/bls-signature-verification.js
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('should not verify a invalid signature', async () => {
+    try {
+      let signatureX =
+        '11181692345848957662074290878138344227085597134981019040735323471731897153462';
+      let signatureY = '12345';
+      let result = await BLS.verify(message, signatureX, signatureY);
```
quiet-node, line 59: We should add a falsy statement here to force the program to go to the `catch` block, e.g.  `expect(true).to.be.false;`. Otherwise in case this line does not throw the error, this UT will pass but not in a way that we expect

test/bls-signature/bls-signature-verification.js
```diff
@@ -0,0 +1,86 @@
...
+    }
+  });
+
+  it('should not verify a invalid message', async () => {
+    try {
+      let signatureX =
+        '11181692345848957662074290878138344227085597134981019040735323471731897153462';
+      let signatureY =
+        '6479746447046570360435714249272776082787932146211764251347798668447381926167';
+      const result = BLS.verify('0x12345', signatureX, signatureY);
```
quiet-node, line 76: We should add a falsy statement here to force the program to go to the `catch` block, e.g.  `expect(true).to.be.false;`. Otherwise in case this line does not throw the error, this UT will pass.

test/bls-signature/bls-signature-verification.js
```diff
@@ -0,0 +1,86 @@
...
+    } catch (err) {
+      console.log(err.message);
```
quiet-node, line 78: Do we need this log?

quiet-node (changes requested):
Good work. I left some suggestions

contracts-abi/contracts/bls-signature/Migrations.sol/Migrations.json
```diff
@@ -0,0 +1,59 @@
+[
```
quiet-node, line 1: you might want to remove this ABI since the contract was removed

contracts/bls-signature/BLS.sol
```diff
@@ -0,0 +1,29 @@
+pragma solidity ^0.8.0;
```
quiet-node, line 1: missing license header

contracts/bls-signature/Pairing.sol
```diff
@@ -0,0 +1,155 @@
+pragma solidity ^0.8.0;
```
quiet-node, line 1: missing license header

quiet-node (changes requested):

quiet-node (approved):
LG but one Q

contracts/bls-signature/Pairing.sol
```diff
@@ -0,0 +1,156 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.0;
+
+/*
+ * @title Pairing
+ * @dev BN128 pairing operations.
+ * @ref https://github.com/JacobEberhardt/ZoKrates/blob/da5b13f845145cf43d555c7741158727ef0018a2/zokrates_core/src/verification.rs
+ */
+
+library Pairing {
```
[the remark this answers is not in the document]
AlfredoG87: Maybe, it would be tricky but possible.
However with BLSTest, all the methods are being indirectly covered.

PR #786, opened 2024-06-01
test: add `debug_traceTransaction` with opcode logger tests

### LINKED ISSUE #783: Add `debug_traceTransaction` with opcode logger tests
### Problem

Recently, the mirror node web3 added a debug opcodes endpoint which is utilized in the relay as well but there is no testing coverage yet.

### Solution

Add `debug_traceTransaction` tests

### Alternatives

_No response_

natanasow:
**Description**:

**Related issue(s)**:

Fixes #783 

**Notes for reviewer**:

**Checklist**


.github/workflows/opcode-logger-testing.yml
```diff
@@ -0,0 +1,50 @@
...
+
+      - name: Upgrade @hashgraph/hedera-local to v2.27.1
+        run: npm install @hashgraph/hedera-local@2.27.1 --save
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@8f1998e9878d786675189ef566a2e4bf24869773 # v1.2.0
+        with:
+          version: nightly
+
+      - name: Run besu node
```
Nana-EC, line 40: Q: why do we need to run besu again.
Is it that the `@OpcodeLogger` tests rely on comparing outputs from besu and local node?

.github/workflows/opcode-logger-testing.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: npm run besu:start
+
+      - name: Run opcode tests against besu
+        run: npx hardhat test --grep "besu comparison" --network besu_local
```
Nana-EC, line 44: Q: What coverage do we get with this test?

package.json
```diff
@@ -28,7 +28,8 @@
     "hh:test": "hardhat test",
     "hedera:start": "npx @hashgraph/hedera-local start --limits=false --dev=true --balance=10000000",
     "hedera:stop": "npx @hashgraph/hedera-local stop",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "besu:start": "docker run -d -v ./utils/besu-configs:/var/lib/besu/ -p 8540:8545 -p 8541:8546 hyperledger/besu:24.6.0 --config-file=/var/lib/besu/customConfigFile.toml"
```
Nana-EC, line 32: Q: Is there a README that should be updated and explains how besu is being used here?

natanasow: I missed adding a README, it's my mistake. Maybe a better idea is to remove this from the npm scripts section and just move the `docker run` command into the `opcode-logger-testing.yml` only? In that way, we'll not confuse the developers with this specific besu testing flow. Or it's better to add a descriptive README for `besu:start` npm script?

.github/workflows/opcode-logger-testing.yml
Nana-EC: Gotcha. I'd lie this called out in some doc somewhere so people understand the flow

package.json
Nana-EC: Add a README to be more descriptive for this flow.
Approach makes sense, just explain it somewhere so it's an easy reference

Nana-EC (dismissed):

Ivo-Yankov (approved):

isavov (approved):

Ivo-Yankov (approved):

PR #886, opened 2024-07-31
chore: add HTS Token Proxy reference contract

### LINKED ISSUE #885: Include HTS Token Proxy contract for reference
In addition to its functional requirements, [HIP-719](https://hips.hedera.com/hip/hip-719) also introduces a call redirect mechanism for tokens created using HTS. Any EVM call to an HTS Token will be redirected to the System Precompile `0x167`. More specifically

> 4. The redirect bytecode is obtained by compiling the following contract, which accepts all the inputs the user provided (including function selector and function arguments), precedes the argument list with the token address, and _delegatecalls_ the HTS precompile:
>
> ```solidity
> // SPDX-License-Identifier: Apache-2.0
> pragma solidity 0.5.5;
> contract Assembly {
>  fallback() external {
>      address precompileAddress = address(0x167);
>      assembly {
>          mstore(0, 0xFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFE)
>          calldatacopy(32, 0, calldatasize())
>          let result := delegatecall(gas(), precompileAddress, 8, add(24, calldatasize()), 0, 0)
>          let size := returndatasize()
>          returndatacopy(0, 0, size)
>          switch result
>              case 0 { revert(0, size) }
>              default { return(0, size) }
>      }
>  }
> }
> ```
> 5. This means that _any_ function can be redirected-to as long as the HTS precompile handles the redirect call.

This bytecode is then returned when an `eth_getCode` call is made on a Token. See https://github.com/hashgraph/hedera-services/blob/fbac99e75c27bf9c70ebc78c5de94a9109ab1851/hedera-node/hedera-smart-contract-service-impl/src/main/java/com/hedera/node/app/service/contract/impl/state/DispatchingEvmFrameState.java#L96.

For future reference and completion, we should include this contract in the repo.

acuarica:
**Description**:

This PR includes the HTS Token Proxy contract for reference.

**Related issue(s)**:

Fixes #885.

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/contracts/hip-719-proxy/HRC719TokenProxy.sol
+++ b/contracts/hip-719-proxy/HRC719TokenProxy.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.4.9 <0.9.0;
+
+/// @dev HTS Token Proxy contract defined in HIP-719, Specification section.
+///
+/// For more information,
+/// see https://github.com/hashgraph/hedera-smart-contracts/issues/885.
+contract HRC719TokenProxy {
+    fallback() external payable {
+        address precompileAddress = address(0x167);
+        assembly {
+            mstore(0, 0xFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFEFE)
+            calldatacopy(32, 0, calldatasize())
+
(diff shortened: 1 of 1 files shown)


Ivo-Yankov (approved):
LGTM

PR #899, opened 2024-08-08
chore: add comments to document revert behavior in view functions

lukelee-sl:
**Description**:
Add @dev comments to document behavior which reverts calls when a HTS view function is not successful.

**Related issue(s)**:

Fixes #8746


natanasow (approved):
LGTM

PR #911, opened 2024-08-19
feat: add new tests for HIP-1010

### LINKED ISSUE #876: HIP-1010: Solidity reference definition
Add the solidity references mentioned in [HIP-1010 design doc](https://github.com/hashgraph/hedera-services/pull/14205/commits)
and include end-user tests also covered in the design document.

konstantinabl:
**Description**:

Testing new functionality regarding fees, namely **_updateFungibleTokenCustomFeesPublic_** and **_updateNonFungibleTokenCustomFeesPublic_** (HIP-1010), but ALSO testing if fees are correctly deducted from accounts, since there were no such test before. Thus, the tests in this PR are slightly bigger and more complex than expected

**Related issue(s)**:

Fixes #876


konstantinabl:
Blocked by https://github.com/hashgraph/hedera-mirror-node/issues/9136 
We need this change in the relay in order to be able to return predefined gas for smart contracts that are in services, but not yet in the mirror node

konstantinabl:
This [issue](https://github.com/hashgraph/hedera-mirror-node/issues/9136) was resolved, however currently waiting for a new release of the local node that will use newest mirror node tag 0.113.1

test/precompile/hedera-token-service/utils.js
```diff
@@ -395,6 +494,64 @@ class Utils {
     return tokenAddress;
   }
 
+  static hexToASCII(str) {
```
natanasow, line 497: There is a similar method defined here https://github.com/hashgraph/hedera-smart-contracts/blob/main/test/precompile/hedera-token-service/redirect-for-token.js#L40. We may consider extracting these methods in utils class?

test/precompile/hedera-token-service/utils.js
```diff
@@ -577,6 +745,18 @@ class Utils {
     }
   }
 
+  static async getAccountBalance(address) {
```
natanasow, line 748: I know it's for test purposes only, but we can use the mirror node for balance querying, right?

natanasow (approved):
LGTM, some nits/questions :rocket:

test/precompile/hedera-token-service/utils.js
konstantinabl: @natanasow yes, we can

PR #986, opened 2024-10-14
ci: Updating one of the actions to fix issues with inputs.

### LINKED ISSUE #957: Manual-testing permissions update with a boolean.
See if there is a way to have it based on some boolean to ask for different permissions inside test-workflow.yml. 
For example if its in the context of a PR to continue asking for:
```
permissions: 
contents: read 
checks: write 
pull-requests: write
```
However, if its not in the context of PR, like a triggered manual job to ask only for:
```
permissions: 
contents: read
```

san-est:
**Description**:
The updated version of the action publish-unit-test-result in the workflow requires new inputs, this PR aims to fix that.
Additionally the attached issue to this PR aimed to see if we can have variable permissions based on how the workflow is run. However, I could not find a suitable solution for the moment and the only action that takes advantage of the extra permissions is publish-test-results so I see no security implications currently.

**Related issue(s)**:

Fixes #957 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/test-workflow.yml
+++ b/.github/workflows/test-workflow.yml
@@ -77,8 +77,8 @@ jobs:
         uses: step-security/publish-unit-test-result-action@4519d7c9f71dd765f8bbb98626268780f23bab28 # v2.17.0
         if: ${{ !cancelled() }}
         with:
-          check_name: '' # Set to empty to disable check run
-          comment_mode: off
+          check_name: 'Unit Tests results' # Set to empty to disable check run
+          comment_mode: on
           json_thousands_separator: ','
           files: 'test-*.xml'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-workflow.yml
```diff
@@ -77,10 +77,10 @@ jobs:
         uses: step-security/publish-unit-test-result-action@4519d7c9f71dd765f8bbb98626268780f23bab28 # v2.17.0
         if: ${{ !cancelled() }}
         with:
-          check_run_disabled: true
+          check_name: '' # Set to empty to disable check run
```
nathanklick, line 80: We should enable the checks. They were previously disabled due to a bug in ARC, but that has now been resolved.

nathanklick (changes requested):

.github/workflows/test-workflow.yml
san-est: I've enabled the checks and it passed.

rbarkerSL (approved):

andrewb1269hg (approved):

PR #1044, opened 2024-12-03
feat: added ContractScanner service class to handle fetching contracts from mirror node

### LINKED ISSUE #1038: [ERC Registry] Add ContractScanner service class to handle fetching contracts from mirror node
### Problem  
According to the design document, a new `ContractScanner` service class is needed to handle fetching contracts from the mirror node.

### Solution  
Implement a `ContractScanner` service class under `tools/erc-repository-indexer/src/services/contractScanner.ts`. The class should include:  
- A `fetchContracts()` method to retrieve all contracts from the mirror node.  
- A `fetchContractByteCode()` method to fetch detailed contract objects using `contractId`.  

The `ContractScanner` class should also be tested using the Jest testing framework.

quiet-node:
**Description**:
This PR introduces the `fetchContracts()` and `fetchContractByteCode()` methods to facilitate the retrieval of contracts from the mirror node. Additionally, it includes test coverage for the associated methods.

**Related issue(s)**:

Fixes #1038

**Notes for reviewer**:

**Checklist**


tools/erc-repository-indexer/erc-contract-indexer/src/utils/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+export default {
+  RETRY_DELAY_MS: 15000,
+  GET_CONTRACT_ENDPOINT: '/api/v1/contracts',
+  MIRROR_NODE_FALL_BACK_BASE_URL: 'https://previewnet.mirrornode.hedera.com',
```
natanasow, line 24: Should we hardcode previewnet here?

tools/erc-repository-indexer/erc-contract-indexer/example.env
```diff
@@ -0,0 +1,3 @@
+HEDERA_NETWORK=
+MIRROR_NODE_URL=
```
natanasow, line 2: Do we need both `HEDERA_NETWORK` and `MIRROR_NODE_URL`? 

If we have a Hedera network, we might be able to create an SDK client via `Client.for<HEDERA_NETWORK>()`. Once we get the client, we will be able to get the mirror node URL by the already initialized client IMO.

natanasow (approved):
LGTM, 2 nits :rocket: .

tools/erc-repository-indexer/erc-contract-indexer/src/utils/constants.ts
quiet-node: Are you referring to the `HEDERA_NETWORK`, or are you discussing the `MIRROR_NODE_FALL_BACK_BASE_URL` being set to previewnet? 

If it's about the `HEDERA_NETWORK`, it may not be necessary and could potentially be removed from the .env file., 

If you're referring to the `MIRROR_NODE_FALL_BACK_BASE_URL` being set to previewnet by default, it's simply a safe fallback, ensuring that the script has a valid URL to revert to.

tools/erc-repository-indexer/erc-contract-indexer/example.env
quiet-node: > Do we need both HEDERA_NETWORK and MIRROR_NODE_URL?

Yes, that's a great question! Currently, the script relies solely on the `MIRROR_NODE_URL` to retrieve contracts, so `HEDERA_NETWORK` could potentially be removed.

> If we have a Hedera network, we might be able to create an SDK client using `Client.for<HEDERA_NETWORK>()`. Once we obtain the client, we can retrieve the mirror node URL through the already initialized client, in my opinion.

That's a valid point! However, I don't see the necessity of using the SDK in this context. We aim to leverage the MN REST server and avoid making calls to the SDK, as those would ultimately route the requests to the mirror node anyway.

quiet-node:
Will update and remove `HEDERA_NETWORK` env variable in future PR.

PR #1057, opened 2024-12-05
feat: added an option for resolveStartingPoint() to fetch next pointer from disk

### LINKED ISSUE #1056: [ERC Registry] add new method to persist the value of next pointer in case interupted
### Problem  
The `/contracts` endpoint is utilized to fetch batches of contracts from the mirror node. Each response object can include a maximum of 100 contracts and relies on pagination to continue the indexing process. Pagination is facilitated by the `next` pointer, which is included in the response object. 

The `next` pointer can be persistently stored, enabling the indexing tool to resume from the stored pointer in case of an interruption during a previous run. Additionally, if the tool completes indexing but new contracts are added later, it can start from the previously stored `next` pointer, avoiding the need to reindex the entire network.

### Solution  
Implement a utility function to store the `next` pointer returned from the mirror node and load it as the starting point configuration.

quiet-node:
**Description**:
This PR enhances `resolveStartingPoint()` by adding the ability to fetch the `next` pointer from disk. To support this, `RegistryGenerator()` has been updated with two new methods: `updateNextPointer()` and `retrieveNextPointer()`. 

The `resolveStartingPoint()` method now determines the starting point for indexing based on configuration settings. It prioritizes the `STARTING_POINT` defined in the configuration; if unavailable, it checks for a stored `next` pointer on disk and uses it if found. If neither is present, indexing defaults to starting from the genesis block.

**Related issue(s)**:

Fixes #1056

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
+++ b/tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
@@ -39,6 +39,13 @@ export class RegistryGenerator {
    */
   private readonly erc721JsonFilePath: string;
 
+  /**
+   * @private
+   * @readonly
+   * @property {string} nextPointerFilePath - The file path where the next pointer for indexing will be stored.
+   */
+  private readonly nextPointerFilePath: string;
+
   constructor() {
     this.erc20JsonFilePath = Helper.buildFilePath(
       constants.ERC_20_JSON_FILE_NAME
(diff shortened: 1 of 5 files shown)


tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
```diff
@@ -39,13 +39,44 @@ export class RegistryGenerator {
...
+    );
+  }
+
+  /**
+   * Updates the next pointer in a file if it is not null.
+   * @param {string | null} next - The next pointer to be written to the file. If null, the file will not be updated.
+   * @returns {Promise<void>} A promise that resolves when the next pointer has been successfully written to the file.
+   */
+  async updateNextPointer(next: string | null): Promise<void> {
+    if (next) {
```
Nana-EC, line 67: How often is this called?
Just curious how often we're writing to file

tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
```diff
@@ -102,38 +136,38 @@ export class RegistryGenerator {
     // Convert Map values back to array for file writing
     const uniqueContracts = Array.from(contractMap.values());
 
-    await this.writeContractsToFile(filePath, uniqueContracts);
+    await this.writeContentsToFile(filePath, uniqueContracts);
+    console.log('Finish pushing new ERC token contracts to registry.');
```
Nana-EC, line 140: nit: we should say how many were written
```suggestion
    console.log('Finished writing x new ERC token contracts to registry.');
```

Nana-EC (approved):
LG, one nit suggestion for a next commit or PR

tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
quiet-node: I plan to write this at the end of each batch of contracts, which is currently set to a maximum of 100. So if there are 1,000 contracts, this will be executed 10 times.

tools/erc-repository-indexer/erc-contract-indexer/src/services/registryGenerator.ts
quiet-node: Gotcha yeah good point! Will update in follow up PRs

PR #1075, opened 2024-12-11
chore: added a list of files containing token information for ERC-20 and ERC-721 tokens on the Hedera networks

### LINKED ISSUE #1074: [ERC Registry] Phase 2: Generate final registry JSON files
### Problem  
The ERC Registry project is now technically complete and can index the Hedera networks to retrieve a list of ERC20 and ERC721 contracts, which are stored in JSON format files. These JSON files need to be checked in so that downstream applications can utilize them.

### Solution  
Execute the ERC Registry project to index the networks, identify ERC contracts (ERC-20 and ERC-721), and record them in the registry. Finally, deploy the code for public visibility.

quiet-node:
**Description**:  
This PR includes a collection of files containing token information for ERC-20 and ERC-721 tokens on the Hedera networks. It represents the final outcome of the ERC Registry project, which aims to index the entire network to identify ERC-20 and ERC-721 token contracts. The project also performs contract calls to the mirror node to retrieve detailed information about these tokens.  

Currently, this PR includes ERC contract data for both testnet and mainnet, up to 10/12/2024.

**Related issue(s)**:

Fixes #1074 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-20.json
+++ b/tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-20.json
@@ -1,0 +1,3122 @@
+[
+  {
+    "contractId": "0.0.14902",
+    "address": "0x0000000000000000000000000000000000003a36",
+    "name": "",
+    "symbol": "",
+    "totalSupply": 0,
+    "decimals": 0
+  },
+  {
+    "contractId": "0.0.15701",
+    "address": "0x0000000000000000000000000000000000003d55",
+    "name": "CUSD",
+    "symbol": "CUSD",
+    "totalSupply": 0,
+    "decimals": 18
+  },
+  {
+    "contractId": "0.0.15707",
+    "address": "0x0000000000000000000000000000000000003d5b",
+    "name": "CUSD",
+    "symbol": "CUSD",
+    "totalSupply": 0,
+    "decimals": 18
+  },
+  {
+    "contractId": "0.0.18586",
+    "address": "0x000000000000000000000000000000000000489a",
+    "name": "CUSD",
+    "symbol": "CUSD",
+    "totalSupply": 0,
+    "decimals": 18
+  },
+  {
+    "contractId": "0.0.20354",
+    "address": "0x0000000000000000000000000000000000004f82",
+    "name": "REJOLUT",
+    "symbol": "REJO",
+    "totalSupply": 1e+27,
+    "decimals": 18
+  },
+  {
+    "contractId": "0.0.21261",
+    "address": "0x000000000000000000000000000000000000530d",
+    "name": "HbarToken",
+    "symbol": "HT",
+    "totalSupply": 10000000,
+    "decimals": 8
+  },
+  {
+    "contractId": "0.0.27482",
+    "address": "0x0000000000000000000000000000000000006b5a",
+    "name": "TOKENWALLET",
+    "symbol": "TW",
+    "totalSupply": 1234567,
+    "decimals": 18
+  },
+  {
+    "contractId": "0.0.27879",
+    "address": "0x0000000000000000000000000000000000006ce7",
+    "name": "TOKENWALLET",
+    "symbol": "TW",
+    "totalSupply": 10,
+    "decimals": 18
+  },
+  {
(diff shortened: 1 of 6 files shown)


tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-20.json
```diff
@@ -0,0 +1,3122 @@
+[
+  {
+    "contractId": "0.0.14902",
+    "address": "0x0000000000000000000000000000000000003a36",
+    "name": "",
```
Nana-EC, line 5: Q: Are the empty name and symbols mistakes?

tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-721.json
```diff
@@ -0,0 +1,110 @@
...
+    "contractId": "0.0.1778110",
+    "address": "0x00000000000000000000000000000000001b21be",
+    "name": "Genesis SpheraHead",
+    "symbol": "SPH"
+  },
+  {
+    "contractId": "0.0.2015187",
+    "address": "0x00000000000000000000000000000000001ebfd3",
+    "name": "UserNftMinter",
+    "symbol": "UMT"
```
Nana-EC, line 12: Q: are we going to add serial ID next or?

Nana-EC (commented):
Looking good, 2 questions

tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-20.json
quiet-node: It appears that the contract was deployed with these fields left uninitialized or empty or whatnot. 

The contract at address 0x0000000000000000000000000000000000003a36 contains bytecode that passes the sevm.isErc() method and successfully responds to all contract mirror node call requests (name, symbol, decimals, totalSupply) but the return values are all empty and 0s.

tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-721.json
quiet-node: Right serial number is a tricky part. In the context of ERC token contracts, the serial number is equivalent to the token ID, and a single token can have multiple token IDs (or serial numbers, if you will). The `name` and `symbol` fields are retrieved via contract calls to the Mirror Node APIs using the standard sighashes for the `name()` and `symbol()` methods. However, I believe there is no standard API to retrieve all token IDs associated with a token.  

Looking at [[this interface from OpenZeppelin](https://docs.openzeppelin.com/contracts/5.x/api/token/erc721#IERC721)](https://docs.openzeppelin.com/contracts/5.x/api/token/erc721#IERC721), the token ID appears to be a predefined variable typically known beforehand and used as a function argument for operations that require it, rather than a retrievable value.  

Even if there's such an API to retrieve all IDs existed, I believe it would be better for downstream applications like MNE to handle this task to avoid outdated information. Just thinking out loud but MNE could address this by making on-demand requests like when a user clicks on a token, the application could fetch a list of owners and their associated token IDs or provide similar functionality. This approach would somehow align with how Etherscan manages on their [token balances](https://etherscan.io/token/0x524cab2ec69124574082676e6f654a18df49a048#balances) pages.

tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-20.json
Nana-EC: Gotcha, I guess we'll leave it for users to verify and use as needed.

tools/erc-repository-indexer/erc-contract-indexer/erc-registry/mainnet/erc-721.json
Nana-EC: Sounds good.
This and the empty fields comment before make me think it'd be good to add a section that talks to known issues that consumers of the file will have to watch out for.
Something to add to the README here or in a follow up PR

Nana-EC (approved):
Nice 1st version

PR #1142, opened 2025-01-06
test: validate contract and admin key in hollow account finalized as contract

### LINKED ISSUE #1007: create test to validate contract admin key in hollow account finalized as contract
### Problem

As of 0.56 of services, hollow account that are finalized as contracts should have a self managed admin key (i.e. key=contract(0.0.x) where x is the contract id of the new contract.)

There is no test to validate this.

### Solution

Write a test that 
- creates a hollow account
- finalizes it as a contract (contract create transaction signed by the private key of the hollow account)
- calls `getContractinfo` to check that the admin key is a contract key of value 0.0.x where x is the hedera contract id number.

### Alternatives

_No response_

nadezhdapopovaa:
**Description**:
Added test for admin key and contract id validation (should be the same) after deploying a contract with hollow account.

**Related issue(s)**:

Fixes #1007

### DIFF
--- a/test/contract-admin-id/contractAdminId.js
+++ b/test/contract-admin-id/contractAdminId.js
@@ -1,3 +1,23 @@
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 const { expect } = require('chai');
 const { ethers } = require('hardhat');
 const utils = require('../system-contracts/hedera-token-service/utils');
--- a/test/contract-admin-id/contractAdminId.js
+++ b/test/contract-admin-id/contractAdminId.js
@@ -1,0 +1,38 @@
+const { expect } = require('chai');
+const { ethers } = require('hardhat');
+const utils = require('../system-contracts/hedera-token-service/utils');
+const Constants = require('../constants');
+
+describe('Admin Key and Contract ID Validation', function () {
+  let signers;
+  let sdkClient;
+  let hollowWallet;
+
+  before(async function () {
+    signers = await ethers.getSigners();
+    sdkClient = await utils.createSDKClient();
+    hollowWallet = ethers.Wallet.createRandom().connect(ethers.provider);
+
+    await (
+      await signers[0].sendTransaction({
+        to: hollowWallet.address,
+        value: ethers.parseEther('100'),
+        gasLimit: 1_000_000,
+      })
+    ).wait();
+  });
+
+  it('should ensure that the admin key matches the contract ID after deploying the contract with the hollow account', async function () {
+    const factory = await ethers.getContractFactory(
+      Constants.Contract.Base,
+      hollowWallet
+    );
+    const contract = await factory.deploy();
+    const info = await utils.getContractInfo(contract.target, sdkClient);
+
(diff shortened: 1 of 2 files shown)


test/contract-admin-id/contractAdminId.js
```diff
@@ -0,0 +1,38 @@
+const { expect } = require('chai');
```
quiet-node, line 1: Let's not forget the license header comment. It should be included in any .js/.ts file.

quiet-node:
Let’s not forget the assignees and additional PR details.

test/contract-admin-id/contractAdminId.js
nadezhdapopovaa: added

quiet-node:
@lukelee-sl please review the PR when you have a chance to see if it aligns with your vision for this feature.

konstantinabl (dismissed):
Good job! LGTM

test/contract-admin-id/contractAdminId.js
```diff
@@ -0,0 +1,58 @@
...
+      hollowWallet
+    );
+    const contract = await factory.deploy();
+    const info = await utils.getContractInfo(contract.target, sdkClient);
+
+    const adminkey = info.adminKey.num;
+    const contractId = info.contractId.num;
+
+    expect(adminkey.equals(contractId)).to.be.true;
+  });
```
lukelee-sl, line 57: Looks good but there is another case we want to ensure coverage of - namely when a contract create happens via an `EthereumTransaction`. for example https://hashscan.io/mainnet/contract/0.0.7546407?pf=1.  Notice that the admin key was set to null.  Any new contracts deployed in this manner should now have the correct self managing key as in the above case

nadezhdapopovaa: we execute the transaction via ethers, which uses the relay, so the transaction is an ethereum transaction.

I ran the test against testnet and we can see (in the links below) that the admin key is the same as the contract id and also in the deploy transaction we can see that the type is ethereum transaction. Or am I wrong?

contract: https://hashscan.io/testnet/contract/0.0.5360643?pf=1&p=1&k=1736330364.039408781
deploy transaction: https://hashscan.io/testnet/transaction/1736330364.039408781

lukelee-sl: I think I was referring to a test that simply deploys a contract (any contract) without necessarily finishing a hollow address as a contract (although I probably expressed this poorly).

natanasow (dismissed):
LGTM

test/contract-admin-id/contractAdminId.js
```diff
@@ -0,0 +1,69 @@
+/*-
+ *
+ * Hedera Smart Contracts
+ *
+ * Copyright (C) 2024 Hedera Hashgraph, LLC
```
natanasow, line 5: nit:
```
 * Copyright (C) 2025 Hedera Hashgraph, LLC
 ```

File: test/contract-admin-id/contractAdminId.js [hunk 4]
```diff
@@ -0,0 +1,69 @@
...
+    const contract = await factory.deploy();
+    const info = await utils.getContractInfo(contract.target, sdkClient);
+
+    const adminkey = info.adminKey.num;
+    const contractId = info.contractId.num;
+
+    expect(adminkey.equals(contractId)).to.be.true;
+  });
+
+  it('should ensure that the admin key matches the contract ID after deploying a contract without finalizing a hollow address', async function () {
```
natanasow, line 59: nit:
```typescript
it('should ensure that the admin key matches the contract ID after deploying a contract', async function() {
```

test/contract-admin-id/contractAdminId.js
```diff
@@ -0,0 +1,69 @@
...
+    const factory = await ethers.getContractFactory(Constants.Contract.Base);
+    const contract = await factory.deploy();
+    const info = await utils.getContractInfo(contract.target, sdkClient);
```
natanasow, line 62: keep in mind that here we don't need to wait for the receipt `await contract.waitForDeployment()` because the next `getContractInfo` call is against services via SDK call but if it was against the mirror node, the data might not be populated yet (if we hit the timeframe of current block)

Ferparishuertas: Shall we add it though?

lukelee-sl (approved):
LGTM.  Thank you very much!

natanasow (approved):

PR #1212, opened 2025-01-28
chore: add schedule service system contract definition files

lukelee-sl:
**Description**:
Add files to define the Hedera Schedule system contract.

Fixes: #1213 

**Checklist**


contracts/system-contracts/hedera-schedule-service/IHRC755ScheduleFacade.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.4.9 <0.9.0;
+pragma experimental ABIEncoderV2;
+
+interface IHRC755 {
+    /// Signs the targeted schedule transaction with the key of the calling EOA.
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    function signSchedule() external returns (int64 responseCode);
+}
```
david-bakin-sl, line 9: itty bitty nit: missing EOL at EOF

david-bakin-sl (approved):
LGTM

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
```diff
@@ -0,0 +1,42 @@
...
+    function authorizeSchedule(address schedule) external returns (int64 responseCode);
+
+    /// Allows for the signing of a schedule transaction given a protobuf encoded signature map
+    /// The message signed by the keys is defined to be the concatenation of the shard, realm, and schedule transaction ID.
+    /// @param schedule the address of the schedule transaction.
+    /// @param signatureMap the protobuf encoded signature map
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    function signSchedule(address schedule, bytes memory signatureMap) external returns (int64 responseCode);
+
+    /// Allows for the creation of a schedule transaction for given a system contract address, abi encoded call data and payer address
```
quiet-node, line 20: nit: Is the word `for` supposed to be there? Or did English just fail me here?

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
```diff
@@ -0,0 +1,42 @@
...
+    /// Currently supports the Hedera Token Service System Contract (0x167) with encoded call data for
+    /// createFungibleToken, createNonFungibleToken, createFungibleTokenWithCustomFees, createNonFungibleTokenWithCustomFees
+    /// and updateToken functions
+    /// @param systemContractAddress the address of the system contract from which to create the schedule transaction
+    /// @param callData the abi encoded call data for the system contract function
+    /// @param payer the address of the account that will pay for the schedule transaction
+    /// @return responseCode The response code for the status of the request. SUCCESS is 22.
+    /// @return scheduleAddress The address of the newly created schedule transaction.
+    function scheduleNative(address systemContractAddress, bytes memory callData, address payer) external returns (int64 responseCode, address scheduleAddress);
```
quiet-node, line 29: I might need to read more about HIP-755 and HIP-756, but in the meantime, could you help clarify the questions below:
	1.	Will there be a scheduling function for regular transactions, or does the scheduling feature only apply to system contract transactions?
	2.	It’s a bit confusing to me that scheduleAddress is used as an identifier for the scheduled transaction. Why did we choose to use a 20-byte address instead of something like a 32-byte transaction hash?

contracts/system-contracts/hedera-schedule-service/IHRC755ScheduleFacade.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity >=0.4.9 <0.9.0;
+pragma experimental ABIEncoderV2;
+
+interface IHRC755 {
```
quiet-node, line 5: Should we use a more specific name like IHRC755ScheduleFacade?

quiet-node (changes requested):

contracts/system-contracts/hedera-schedule-service/IHRC755ScheduleFacade.sol
lukelee-sl: fixed

contracts/system-contracts/hedera-schedule-service/README.md
```diff
@@ -0,0 +1,21 @@
+## Hedera Account Service (HAS) System Contract Functions
```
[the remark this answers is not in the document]
lukelee-sl: Yikes.  thanks. fixed

contracts/system-contracts/hedera-schedule-service/README.md
```diff
@@ -0,0 +1,21 @@
...
+| `getScheduledCreateNonFungibleTokenInfo`    | `0xd68c902c`           | [0.59](https://docs.hedera.com/hedera/networks/release-notes/services#release-v0.59) | [HIP 756](https://hips.hedera.com/hip/hip-756) | `getScheduledCreateNonFungibleTokenInfo(address scheduleAddress) external returns (int64 responseCode, IHederaTokenService.NonFungibleTokenInfo memory nonFungibleTokenInfo)` |
+
+The Hedera network also make facade contract calls available to EOAs for improved experience.
+Facade function allow for EOAs to make calls without requiring a deployed contract
+The table below outlines the available Hedera Schedule Service (HSS) System Contract facade functions:
+
+| Function Name                          | Function Selector Hash | Consensus Node Release Version                                               | HIP                                            | Method Interface                                                                                                                                                              | Comments                                                       |
```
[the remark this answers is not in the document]
lukelee-sl: removed.

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
lukelee-sl: I think it was in the wrong place

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
lukelee-sl: 1. only certain system contract functions are allowed at present - Create{Fungible,NonFungible}Token{WithCustomFees} and UpdateTokenInfo
2. a `Schedule` is an entity similar to `Account` and `Token` in Hedera.  We are using the same long zero encoding of  entity id for `Schedule` to be consistent.

contracts/system-contracts/hedera-schedule-service/IHRC755ScheduleFacade.sol
lukelee-sl: yes, fixed

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
quiet-node: Ah, so scheduleAddress represents the long zero encoding of the token ID 0.x.y, correct? Then what’s the logic behind calculating the token ID? To my understanding, a scheduled transaction is one that can be executed in the future, right? How does it determine its token ID at the time scheduling the transaction? Don’t token IDs increment sequentially as new tokens are added?

quiet-node (approved):
LGTM! Great work!

david-bakin-sl (approved):
LGTM

contracts/system-contracts/hedera-schedule-service/IHederaScheduleService.sol
lukelee-sl: > Ah, so scheduleAddress represents the long zero encoding of the token ID 0.x.y, correct
No actually, the the newly created scheduled transaction itself has a 0.0.<schedule_num> id that is create and we must refer to address in order to tell which scheduled transaction is which.
There is a 0.0.<token_num> id that is also "reserved" for the prospective token which is about to be but not yet create token in the scheduled transaction.

PR #1265, opened 2025-02-27
feat: extend test coverage on `WHBAR` contract

### LINKED ISSUE #1264: Extend test coverage on `WHBAR` contract
### Problem

Currently, we have 100% line coverage on WHBAR but that doesn't mean we've covered all of the business logic test cases.

### Solution

Extend the current test plan.

Test plan in notion

### Alternatives

_No response_

natanasow:
### Problem

Currently, we have 100% line coverage on WHBAR but that doesn't mean we've covered all of the business logic test cases.

### Solution

Extend the current test plan.

https://limechain.notion.site/WHBAR-Test-Plan-1a610baa8dcd8062ad17d436fb5a1627?pvs=4

**Related issue(s)**:

Fixes #1264 

**Notes for reviewer**:
It seems Foundry nightly runs into the following issue

`forge: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by forge)`

To solve this, we instead switch to stable version.

**Checklist**


test/wrapped-tokens/WHBAR.js
```diff
@@ -172,97 +270,523 @@ describe('WHBAR', function() {
...
+      expect(error.message).to.include('value out-of-bounds');
+    }
+
+    // Test with MAX_UINT256 which should work (no overflow in approve)
+    await expect(contract.approve(spender, MAX_UINT256)).not.to.be.reverted;
+  });
+
+  it('WHBAR-018 should revert on negative value for deposit', async function () {
+    try {
+      await contract.deposit({ value: '-1' });
```
konstantinabl, line 487: Wouldn't it be better if we utilize chaiAsPromised to make the test more readable, instead of using a try/catch. For me its unecessary
What I suggest is something like
`await expect(contract.transferFrom(sender, receiver, OVERFLOW_VALUE)).to.be.rejectedWith(error)`

konstantinabl (changes requested):
left a suggestion : )

test/wrapped-tokens/WHBAR.js
simzzz: I agree, that would simplify it. I will get it done.

konstantinabl (approved):

PavelSBorisov (approved):

PR #1267, opened 2025-02-27
chore: add whbar tests pipeline

### LINKED ISSUE #1266: Add `WHBAR` contract tests to the CI
### Problem

Currently, the WHBAR tests are not running within the CI.

### Solution

Add WHBAR tests to the CI.

### Alternatives

_No response_

natanasow:
**Description**:

Currently, the WHBAR tests are not running within the CI.

**Solution**:

Add WHBAR tests to the CI.
**Related issue(s)**:

Fixes #1266

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -129,6 +129,12 @@ jobs:
     with:
       testfilter: HIP904
 
+  WHBAR:
+    name: WHBAR Contract Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: WHBAR
+
   PublishResults:
     name: Publish Results
     if: ${{ !cancelled() }}
@@ -151,6 +157,7 @@ jobs:
       - BLSSignatureVerification
       - HederaAccountService
       - HIP904
+      - WHBAR
 
     runs-on: smart-contracts-linux-large
     steps:


.github/workflows/tests.yml
```diff
@@ -129,6 +129,12 @@ jobs:
     with:
       testfilter: HIP904
 
+  WHBAR:
+    name: WHBAR Contract Test Suite
+    uses: ./.github/workflows/test-workflow.yml
+    with:
+      testfilter: WHBAR
```
Ferparishuertas, line 136: I suggest to only execute it of changes on the given test or src folder for WHBAR

natanasow:
@hashgraph/release-engineering-managers kindly pinging for review.

quiet-node (approved):

natanasow:
@hashgraph/release-engineering-managers pinging for review :pray:.

.github/workflows/tests.yml
natanasow: I don't think we have that kind of mechanism yet.

We have a base flow, which is defined [here](https://github.com/hashgraph/hedera-smart-contracts/blob/main/.github/workflows/test-workflow.yml). And here is [the list](https://github.com/hashgraph/hedera-smart-contracts/blob/main/.github/workflows/tests.yml#L23) of all jobs. They are executed no matter if there is a change or not.

We may open a new task for making the job execution mandatory only if there is a change. The change must be done on a global level, not only for `WHBAR` IMO.

nathanklick (approved):

PR #1295, opened 2025-03-27
feat: remove skip flag from tests

### LINKED ISSUE #1208: Remove skipped tests for invalid token addresses and more than 10 airdrops in one token transfer list
Currently some tests specifically for airdrop cases such as trying to claim/cancel an airdrop with an invalid token address, or when the token transfer list contains more than 10 tokens (which is the limit) are skipped. The reason for this is that they revert with `INVALID_OPERATION_ID` which is a generic exception and should be `INVALID_TOKEN_ID` for example. I originally thought this exception was the correct response, and since I can't catch it with `getHTSResponseCode()`, I just asserted for a reverted transaction. After more careful analysis this approach is not ideal and they should be changed to assert for the correct message once it's fixed on the services side.

simzzz:
**Description**:
This PR removes the skip flag from the remaining HIP-904 tests. It also includes an additional utility function to parse the response message for better readability for them.

**Related issue(s)**:

Fixes #1208


test/system-contracts/hedera-account-service/ihrc-632/aliasAccountUtility.js
```diff
@@ -94,7 +94,7 @@ describe('@HAS IHRC-632 Test Suite', () => {
       ); // evm address
     });
 
-    it('Should execute getEvmAddressAliasPublic with not long zero address and get INVALID_ACOUNT_ID', async () => {
```
simzzz, line 97: This test had been verifying incorrect behavior which has been fixed here - https://github.com/hiero-ledger/hiero-consensus-node/issues/17570 and was therefore failing after updating the image tag

PR #1359, opened 2025-05-29
test: Remove code duplicates (#1199)

### LINKED ISSUE #1199: Remove repetitive deploy functions from hedera-token-service/utils.js and use the generic deployContract()
### Problem

Currently in hedera-token-service/utils.js there are repeated functions that deploy a contract, but all the logic is the same. There also exists a generic deployContract() function which accepts the contract as a parameter. This adds unnecessary code repetition and makes the code harder to maintain.

### Solution

- Remove all functions that deploy a contract in hedera-token-service/utils.js except the generic one deployContract
- Replace their usage across the codebase with deployContract and pass the contract as a parameter instead 

### Alternatives

_No response_

arianejasuwienas:
**Description**:

Removing duplicated code fragment in the hedera token service smart contract tests.

**Related issue(s)**:

Fixes #1199

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/test/system-contracts/hedera-token-service/utils.js
+++ b/test/system-contracts/hedera-token-service/utils.js
@@ -65,157 +65,61 @@ class Utils {
   }
 
   static async deployERC20Mock() {
-    const erc20MockFactory = await ethers.getContractFactory(
-      Constants.Path.HIP583_ERC20Mock
-    );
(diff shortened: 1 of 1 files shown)


quiet-node (approved):
Good work!

PR #1709, opened 2026-06-03
chore: bump solo to 0.74.0 and remove consensus port-forward workaround

simzzz:
## What & why

Solo's one-shot `falcon deploy` previously did not forward the consensus gRPC port to the host by default, so the Hiero JS SDK timed out. The repo worked around this with `--force-port-forward` flags plus a manual `kubectl port-forward` of `50211`. Solo **0.74.0** fixes the port-forward logic so consensus gRPC (`35211`), JSON-RPC relay (`37546`) and mirror REST ingress (`38081`) are forwarded to the host **by default**, making the workaround unnecessary.

Closes #1683 

## Verification

Deployed Solo 0.74.0 locally and confirmed, with **neither** the `--force-port-forward` flags **nor** the manual `kubectl port-forward`:

- Consensus gRPC reachable on host `35211` (SDK `AccountBalanceQuery` succeeded)
- JSON-RPC relay reachable on `37546` (chainId 298)

The exposed consensus host port is unchanged (`35211` → pod `50211`), so `utils/constants.js` and the manual `createSDKClient` remain correct and are intentionally left as-is.

## Changes

- Bump `@hashgraph/solo` `0.72.0` → `0.74.0` (`package.json`, `setup-solo-stack.sh`, the test / erc-registry / opcode-logger / migration workflows, and docs)
- Drop the redundant `--force-port-forward: true` flags from `falcon.yml` and `setup-solo-stack.sh`
- Trim `solo-kubectl-port-forward.sh` to only the mirror Web3 (`8545`) forward still needed by the ERC-registry indexer; remove the dead consensus `50211` forward
- Remove the now-redundant port-forward CI step from the `test`, `opcode-logger` and `migration` workflows; keep it (renamed) only in `erc-registry-test.yml` where `8545` is used
- Update the indexer README and workflows documentation accordingly

## Notes

- `solo-migration-inplace-upgrade.sh` keeps its `--force-port-forward` flags on purpose — those re-establish forwards after the mid-workflow pod restart, a different subcommand context not covered by the one-shot default behavior.
- The pre-existing prettier debt in `.github/*.yml` (double vs single quotes) was left untouched to avoid unrelated churn.

### DIFF
--- a/.github/falcon.yml
+++ b/.github/falcon.yml
@@ -2,16 +2,12 @@ network:
   --release-tag: "v0.72.0"
 setup:
   --release-tag: "v0.72.0"
-consensusNode:
-  --force-port-forward: true
 mirrorNode:
   --enable-ingress: true
   --pinger: true
   --mirror-node-version: "v0.151.0"
   --values-file: .github/mirror-node-values.yaml
-  --force-port-forward: true
 relayNode:
   --relay-release: "0.76.2"
   --node-aliases: "node1"
   --values-file: .github/relay-values.yaml
-  --force-port-forward: true
--- a/.github/workflows/erc-registry-test.yml
+++ b/.github/workflows/erc-registry-test.yml
@@ -60,10 +60,10 @@ jobs:
         id: solo_stack
         run: bash .github/workflows/support/scripts/setup-solo-stack.sh
         env:
-          SOLO_VERSION: "0.72.0"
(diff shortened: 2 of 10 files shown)


quiet-node (approved):
LGTM

andrewb1269 (approved):

